PR #5, opened 2023-01-30
copy from distributed to local when local is a nullptr

haichangsi:


rscohn2:
I don't like duplicating functions so we can accept nullptr and prefer making the template accept any type. Stopping there would be OK, but it is also possible to define a concept that accepts nullptr and contiguous_iterator

lslusarczyk:
> I don't like duplicating functions so we can accept nullptr and prefer making the template accept any type. Stopping there would be OK, but it is also possible to define a concept that accepts nullptr and contiguous_iterator

I implemented this. Could you please have a look and confirm if this is what you was thinking about?

PR #6, opened 2023-01-30
remove direct references to std::ranges

rscohn2:


rscohn2:
@BenBrock This commit breaks shp build. I did not include shp in CI because it is  difficult to install the compiler on a github CI runner.

One problem is that concepts/concepts.hpp conflicts with a filename in ranges-v3. Renaming fixes that problem, but then there are other issues that I do not know how to resolve. Is this something you could look at?

PR #8, opened 2023-01-31
Add function to transpose from a local vector to distributed one and vice versa

KFilipek:
This PR introduces two functions to transpose, where one argument is a local data structure.

### DIFF
--- a/test/gtest/cpu/mdspan.cpp
+++ b/test/gtest/cpu/mdspan.cpp
@@ -192,9 +192,7 @@ TEST(CpuMpiTests, distribute_to_local_transpose) {
   }
 
   if (comm_rank == 0) {
-    EXPECT_EQ(dsrc(0, 0), ldst(0, 0));
-    EXPECT_EQ(dsrc(1, 0), ldst(0, 1));
-    EXPECT_EQ(dsrc(rows - 1, cols - 2), ldst(cols - 2, rows - 1));
+    expect_eq(dsrc, ldst, true);
   }
 }
 
@@ -219,8 +217,6 @@ TEST(CpuMpiTests, local_to_distribute_transpose) {
   }
 
   if (comm_rank == 0) {
-    EXPECT_EQ(lsrc(0, 0), ddst(0, 0));
-    EXPECT_EQ(lsrc(1, 0), ddst(0, 1));
-    EXPECT_EQ(lsrc(rows - 1, cols - 2), ddst(cols - 2, rows - 1));
+    expect_eq(lsrc, ddst, true);
   }
 }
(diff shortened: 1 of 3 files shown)


rscohn2:
CTest logs are saved as artifacts. The fails are:
```
[  FAILED  ] CpuMpiTests.local_to_distribute_transpose
```

KFilipek:
@rscohn2 @lslusarczyk this PR is ready to review

test/gtest/cpu/mdspan.cpp
```diff
@@ -167,3 +169,58 @@ TEST(CpuMpiTests, DISABLED_distributed_mdarray_transpose) {
...
+    lib::collective::transpose(0, std::make_optional(lsrc), ddst);
+  } else {
+    lib::collective::transpose(
+        0, std::optional<std::experimental::mdspan<T, dyn_2d>>(), ddst);
+  }
+
+  if (comm_rank == 0) {
+    EXPECT_EQ(lsrc(0, 0), v[0]);
+    EXPECT_EQ(lsrc(1, 0), ddst(0, 1));
+    EXPECT_EQ(lsrc(rows - 1, cols - 2), ddst(cols - 2, rows - 1));
```
lslusarczyk, line 224: Can we try to use `check_eq` method here? It checks all items in the matrix. Currently it is executed on local mdspans only but I guess it should run well on distributed too.

lslusarczyk (changes requested):
A few comments to @KFilipek.
I'd like to try a one small refactor related to `optionals` here, so @rscohn2 please hold with review for a while.

test/gtest/cpu/mdspan.cpp
KFilipek: I've added this change as a separate commit

rscohn2:
> I'd like to try a one small refactor related to `optionals` here, so @rscohn2 please hold with review for a while.

Not sure if you are done. It looks ok to me. The optional seems better than nullptr when it fits. I am thinking that we should not support nullptr as a direct argument. User can assign nullptr to an appropriately typed pointer that is passed. It would simplify the types and be more natural SPMD style for me.

rscohn2:
In the near term, we just need something that works for transpose and continuing with `mdspan` is ok.

lslusarczyk:
@rscohn2 , please take a look at https://github.com/oneapi-src/distributed-ranges/pull/24 PR, which consists of this changes + remove of optional, as mdspan can be empty itself. If it looks good to you let's close this PR and merge that PR.

rscohn2:
Superseded by #24

PR #9, opened 2023-01-31
[CI] Append output from ctest failure

KFilipek:
After this change ctest should print output what went wrong during test execution.


rscohn2:
Full ctest logs are saved as artifacts. This change will put errors in the log, which will be more convenient.

PR #10, opened 2023-01-31
add multi-process model compatible with shp

rscohn2:


rscohn2:
@BenBrock: This PR is getting large and I would like to merge. I can merge and fix issues later if you are occupied with other work. There were mostly small changes to SHP. I like the way it came out. The mhp distributed_vector and cpu_algorithms.hpp are simple.

PR #4, opened 2023-01-26
Implement copy when size is not known

haichangsi:
Implement a collective copy without providing the size of the source in the destination.

PR #11, opened 2023-02-01
copy operations final cleanups

lslusarczyk:
I tried to clean up all the stuff related to copy
- functions are in SPMD style (one function accepting either iterators or nullptr)
- functions have consistent signatures ('concept auto name' whenever possible instead of template, unless we need the same type for two params)
- functions have short consistent comments

Maybe it is not best to review the change itself (or not only to review the change), but to look how all copy functions currently look and if we like how do they look.

PR is based on https://github.com/oneapi-src/distributed-ranges/pull/4


rscohn2 (approved):
Looks good. Thanks.

PR #15, opened 2023-02-02
Update `shp` to work with range-v3

BenBrock:


BenBrock:
@rscohn2 This PR is ready to merge.  I made a few changes to shared code in segments tools and also added a ranges shim header (linked comments below).  If you would, take a quick look at those changes and then merge this PR if everything looks reasonable to you.

include/dr/details/device_subrange.hpp
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include <dr/concepts/concepts.hpp>
+#include <dr/details/ranges_shim.hpp>
+#include <iterator>
+
+namespace lib {
+
+template <std::forward_iterator I>
+class device_subrange : public rng::subrange<I, I> {
```
BenBrock, line 14: I added `device_subrange` as a generic subrange we can use inside `segments_tools` while preserving a remote range's rank.  We should perhaps rename this to `remote_subrange`.

include/dr/details/enumerate.hpp
```diff
@@ -0,0 +1,59 @@
...
+    return enumerate(std::forward<R>(r));
+  }
+
+  template <rng::viewable_range R>
+  friend auto operator|(R &&r, const enumerate_adapter_closure &closure) {
+    return enumerate(std::forward<R>(r));
+  }
+};
+
+inline auto enumerate() { return enumerate_adapter_closure(); }
```
BenBrock, line 55: I'm currently using this enumerate implementation.  It seems to work reasonably well.  However, there is a version of enumerate shipping in [C++23](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2164r9.pdf), and also one in range-v3 I haven't looked at.  We should eventually switch to those long-term.

include/dr/details/ranges_shim.hpp
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+// TODO: use libstdc++ 13.0 or greater if available.
+
+// #define DR_USE_STD_RANGES
```
BenBrock, line 9: I haven't figured out the correct macros yet, but we should be able to automatically include standard ranges if we detect that the standard library is recent enough.

include/dr/details/segments_tools.hpp
```diff
@@ -4,52 +4,93 @@
...
 namespace lib {
 
 namespace internal {
 
-// return number of full segments and remainder to cover n elements
+// count the number of segments necessary to cover n elements,
+// returning the index of the last segment and its remainder
 template <typename R>
-void n_segs_remainder(R &&segments, std::size_t n, auto &n_segs,
+void n_segs_remainder(R &&segments, std::size_t n, auto &last_seg,
```
BenBrock, line 20: I made a few changes to segments tools to avoid leaving an empty segment at the end each time as well as to use the new distributed iterator design.

include/dr/details/enumerate.hpp
lslusarczyk: Indeed there is an enumerate in range-v3. I've already removed one implementation of enumerate from our codebase. I propose to use already implemented stuff by default, unless we know we need something different.

include/dr/details/ranges_shim.hpp
rscohn2: I would like to be able to force it from CMAKE, which would require `ifdef` control. We can also support auto selection.

rscohn2 (approved):

rscohn2:
MHP uses same iterator for distributed and remote ranges. SHP uses same iterator for remote and local ranges. MHP assumes `&`distributed reference is a distributed iterator. Some of the code to support views will not be sharable. mhp::distributed_vector and shp::distributed_vector types are different so it should be possible to handle without ifdefs.

rscohn2:
We need to see a lot more application code before we will know how the difference between the 2 models will work out.

include/dr/details/device_subrange.hpp
BenBrock: Sounds good, I will go change it.

include/dr/details/enumerate.hpp
BenBrock: There's an enumerate in range-v3, but not in current implementations of the standard library, since it was just added to C++23, so this needs to stay in for now.

include/dr/details/ranges_shim.hpp
BenBrock: Sounds good, so we'll have maybe a `DR_USE_STD_RANGES` and a `DR_USE_RANGE_V3` that can be set, and if neither of those are set we'll select automatically?

lslusarczyk: Having auto selection only (no flag setting in cmake) will limit test cases we need to support (e.g. gcc-13 with v3-ranges will not be needed to be tested).

BenBrock:
> MHP uses same iterator for distributed and remote ranges. SHP uses same iterator for remote and local ranges. MHP assumes `&`distributed reference is a distributed iterator. Some of the code to support views will not be sharable. mhp::distributed_vector and shp::distributed_vector types are different so it should be possible to handle without ifdefs.

I understand this will affect application code, particularly stencils.  How much of our actual shared views/algorithms do you think will be impacted?  Seems to me like most views/algorithms should be implementable without needing distributed reference's reference operator to return a distributed iterator?

include/dr/details/ranges_shim.hpp
rscohn2: We can decide not to support some combinations, like gcc-13 + range-v3, but I still want to be able to configure it that way for isolating issues.

rscohn2:
zip might be only view/algorithm that doesn't have a `base`. Custom zip might be ok and you already implemented it. I am more worried about maintaining a lot of code for many views than the potential interoperability issue of the types being different.

I don't know if restricting application code is a problem.  We can see how it goes.

PR #18, opened 2023-02-06
Add template overloads for transpose

KFilipek:
Following PR introduces the possibility to easily pass `std::nullopt` for `transpose()` as substitution of `std::optional<T, DExtents>`.
The generic template requires that the underlying type of `mdspan` and `mdarray` is the same and prevents from passing `std::nullopt`.

Disclaimer:
Depends on #8


rscohn2:
Superseded by another PR

Issue #14, opened 2023-02-02
Implement `shp::copy_async` for local to distributed copying


BenBrock:
In my SYCL code base, I've implemented two versions of `copy` for use with distributed ranges and remote ranges.  One conforms to the [standard library's API](https://en.cppreference.com/w/cpp/algorithm/copy), and the other is an asynchronous version that returns a [SYCL event](https://oneapi-src.github.io/DPCPP_Reference/iface/event.html).  The asynchronous version, `copy_async`, returns immediately, and will only ensure the copy is complete when you call `event.wait()` on the SYCL event returned from the function.

I've been filling in implementations of `[shp::copy](https://github.com/intel-sandbox/libraries.runtimes.hpc.dds.distributed-ranges/blob/main/include/dr/shp/copy.hpp)` as I need them.  Currently they're mostly for [copying from `device_ptr`'s](https://github.com/intel-sandbox/libraries.runtimes.hpc.dds.distributed-ranges/blob/main/include/dr/shp/copy.hpp#L44) to local memory and vice-versa.  (Device pointers point to locations in SYCL USM device memory.)

I also recently implemented versions of `shp::copy` that work with distributed ranges.  This means you can copy from a distributed range (split up amongst multiple GPUs) to a local range, or vice-versa.  Take a look at the [distributed -> local version of `copy_async`](https://github.com/intel-sandbox/libraries.runtimes.hpc.dds.distributed-ranges/blob/main/include/dr/shp/copy.hpp#L110).  We can implement this using the `segments` CPO, which returns all the segments in the distributed range our distributed iterator is pointing to, as well as the `segment_index` and `local_index` CPOs, which tell us exactly where our distributed iterator is pointing (to which segment and which index within that segment).  When then go through and [issue remote -> local copies](https://github.com/intel-sandbox/libraries.runtimes.hpc.dds.distributed-ranges/blob/main/include/dr/shp/copy.hpp#L123) for each segment that needs to be copied over.

I went ahead and implemented both a `copy()` for both distributed -> local and local -> distributed.  Currently, however, I only have `copy_async` for distributed -> local.

[...]

PR #22, opened 2023-02-08
shp::copy_async() local -> distributed + tests in both directions

mateuszpn:


KFilipek:
Please remove the extra line in include/dr/shp/copy.hpp:147

include/dr/shp/copy.hpp
```diff
@@ -105,6 +105,48 @@ OutputIt copy(InputIt first, InputIt last, OutputIt d_first) {
...
+    total_copied += n_to_copy;
+  }
+
+  sycl::queue q;
+
+  auto root_event = q.submit([=](auto &&h) { h.depends_on(events); });
+
+  return root_event;
+}
+
```
KFilipek, line 148: Remove extra line

lslusarczyk:
closes #14

.pre-commit-config.yaml
```diff
@@ -41,9 +41,17 @@ repos:
   - id: doc8
 
 - repo: https://github.com/pycqa/isort
-  rev: 5.10.1
+  rev: 5.12.0
   hooks:
   - id: isort
+    args: ['--order-by-type', '--length-sort', "--profile", "black", "--filter-files"]
```
lslusarczyk, line 47: I understand it was important to use higher version of isort in order to make code running on gdds.

Are below changes also needed? If they are not needed please remove.

include/dr/shp/copy.hpp
```diff
@@ -105,6 +105,47 @@ OutputIt copy(InputIt first, InputIt last, OutputIt d_first) {
   return rv;
 }
 
+template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
+cl::sycl::event copy_async(InputIt first, InputIt last, OutputIt d_first) {
```
lslusarczyk, line 109: Once https://github.com/oneapi-src/distributed-ranges/pull/15 is merged please merge your branch to the latest main and add a test for this new _copy_async_ code.

lslusarczyk (changes requested):
Unittest needed

.pre-commit-config.yaml
mateuszpn: Verified as not needed - removed

lslusarczyk (approved):

include/dr/shp/copy.hpp
```diff
@@ -13,20 +13,22 @@
 
 namespace shp {
 
+template <typename Src, typename Dest>
+concept is_syclmemcopyable = std::is_same_v<std::remove_const_t<Src>, Dest> &&
```
BenBrock, line 17: I recommend putting this inside an anonymous namespace.

```cpp
namespace {
}
```

BenBrock: We should not be defining any symbols except for `inline` symbols, but I'm also fine with `__detail`.

rscohn2: > We should not be defining any symbols except for `inline` symbols, but I'm also fine with `__detail`.

Duplication can also come from putting classes in anonymous namespaces. The implementations of the methods will not be shared across compilation units.

include/dr/shp/copy.hpp
```diff
@@ -91,29 +74,67 @@ OutputIt copy(InputIt first, InputIt last, OutputIt d_first) {
...
+    auto &&segment = segments[segment_id];
+    std::size_t n_in_segment = segment.size();
+
+    std::size_t n_to_copy =
+        std::min<size_t>(n_in_segment, std::distance(first, last));
+
+    auto local_last = first;
+    std::advance(local_last, n_to_copy);
+
+    auto remote_iter = segment.begin();
```
lslusarczyk, line 118: `remote_iter` variable is used in one place only. There is no point in making this variable. Just pass `segment.begin()` into `shp::copy_async` call (here and in the function above)

test/gtest/shp/copy.cpp
```diff
@@ -4,42 +4,144 @@
...
+  using LocalVec = std::vector<typename AllocT::value_type>;
+};
 
-TEST(ShpTests, Copy_Dist2Local) {
-  const int n = 100;
-  std::size_t n_to_copy = 20;
+using AllocatorTypes =
+    ::testing::Types<shp::device_allocator<int>,
+                     shp::shared_allocator<long long unsigned int>>;
+TYPED_TEST_SUITE(CopyTest, AllocatorTypes);
```
rscohn2, line 16: This is nice. Can we do share tests for `shp::distributed_vector` and `mhp::distributed_vector` this way?

lslusarczyk: Yes, we can.
But let's do it in separate PR as it is not yet obvious to me how to do it in best way.

PR #26, opened 2023-02-10
mhp support for rng::zip_view

rscohn2:
Allow parallel algorithms on `zip_view` that contain `distributed_range`. Uses CPO's to provide `segments` and friends. Adding support for `transform_view` should leverage the same code.

I will handle zips that contain iota and misaligned distributed vectors in a separate PR.

@lslusarczyk and @BenBrock: Feedback is welcome.


rscohn2:
Merging, will address any review comments after.

PR #31, opened 2023-02-14
conformant changed to aligned, and generalize support

rscohn2:
distributed ranges `a` & `b` are aligned if `rank(a[i]) == rank(b[i])` for all `i`
If `a` and `b` are aligned, then you can do a distributed `for_each` without any communication.

Check the segments instead of individual elements. If you zip together unaligned distributed ranges, then the segments are empty. SHP will probably want different behavior.


include/dr/concepts/concepts.hpp
```diff
@@ -39,8 +39,6 @@ concept distributed_iterator =
     std::forward_iterator<I> &&
     requires(I &iter) {
       { lib::ranges::segments(iter) } -> rng::forward_range;
-      { lib::ranges::segment_index(iter) };
-      { lib::ranges::local_index(iter) };
```
rscohn2, line 43: @BenBrock: I removed these so I could finish the PR. I did not remove all the code for `segment_index`.

PR #43, opened 2023-02-20
move transform view to common

rscohn2:
* move transform view to common
* add a test for it in mhp (currently disabled because it does not compile)
* add shp dot_product build to cmake so there will be an shp test for it
* move some includes to shp.hpp so it will be easier to move header files


include/dr/views/transform.hpp
```diff
@@ -23,12 +20,12 @@ class transform_view : public rng::view_interface<transform_view<V, F>> {
   auto end() const {
     auto segs = segments();
     return normal_distributed_iterator<decltype(segments())>(
-        segs, size_t(segs.size()), 0);
+        segs, std::size_t(segs.size()), 0);
```
rscohn2, line 23: @BenBrock: Some views do not have a size. Also, shouldn't the last argument be the size of the last segment?

BenBrock: Can we guarantee that the segments view itself will have a size, even if all the individual segments themselves don't?

(i.e. can we say the thing you get back from `segments()` is a `sized_range`?)

If not, I think we can at least say it's a `forward_range`, and then we can check `std::distance(first, last)`, although I think we can probably say it's a sized range?

As for the end iterator, the way it's implemented, the `segment_index` will be incremented as soon as you reach the end of a segment, including the last segment, so the iterator you get at one past the final element is actually the 0'th element of one past the last segment.

rscohn2: I don't remember whether it was the segments or the segment which did not have a size. I don't understand why my mhp segments don't have a size. It looks like there is zip view with an unbounded iota inside. I was already planning to reimplement distributed_vector so it directly provides segments and relies on normal_distributed_iterator to provide the flat view, so I will do that first. Also, we should reimplement the other view code to be like transform_view. After that, the problem might go away. Or at least we would have more common code between shp and mhp.

include/dr/views/transform.hpp
```diff
@@ -23,12 +20,12 @@ class transform_view : public rng::view_interface<transform_view<V, F>> {
...
   }
 
   auto segments() const {
     return lib::ranges::segments(base_) |
-           rng::views::transform([=](auto &&segment) {
+           rng::views::transform([this](auto &&segment) {
```
BenBrock, line 28: I suggest updating the capture to `[*this]` or `[fn_]`.  Technically I think we want to copy `fn_` by value, since that's how it's done everywhere in the public-facing API in the standard, and it's required to be copy constructible.

BenBrock: On second thought, we probably really want `=` or `fn_`, since `*this` would also copy the segments by value.  (In theory this should be cheap, but for some of our `segments()` that are vectors with a bunch of stuff, there's no reason to require copying this when we don't need to.)

rscohn2: > we probably really want `=` or `fn_`, since `*this`

Neither are accepted by gcc12:
```
 error: implicit capture of 'this' via '[=]' is \
deprecated in C++20 [-Werror=deprecated]
   28 |            rng::views::transform([=](auto &&segment) {
      |                                  ^
```
and
```
/nfs/site/home/rscohn1/local/projects/dds/dr/include/./dr/views/transform.hpp:28:35: error: capture of non-variable 'lib::transform_\
view<V, F>::fn_'
   28 |            rng::views::transform([fn_](auto &&segment) {
      |                                   ^~~
```
This captures the value and compiles without error.
```
auto segments() const {
    auto fn = fn_;
    return lib::ranges::segments(base_) |
           rng::views::transform([fn](auto &&segment) {                                                                                           return segment | rng::views::transform(fn);
           });
  }
```

BenBrock: Ah, thanks for catching this... I've run into a couple of cases recently where Clang's front-end lets through things that GCC rejects.  Looks like you found a good fix.

PR #44, opened 2023-02-21
For each simplified

lslusarczyk:
* removed `ExecutionPolicy` template as it is always a `device_policy` and there is no performance gain in allowing it to be both: r-value-ref and const ref, using const-ref. On the other hand this template obfuscated code a lot.
* changed `auto && devices` into simple `std::span<const cl::sycl::device> devices`. span itself is a view, there is no gain in making it a reference and it is more readable to see the type itself in the code
* removed `forward` in `std::forward<ExecutionPolicy>(policy).get_devices()` - there is no point in having forward as 'get_devices() method is not specialized for r-value and l-value references
* removed call to `local` in `lib::ranges::local(rng::begin(segment))` as for_each can be provided with e.g. output of `zip_view` or `enumerate` and local on such a pair may be both not avail and to have no sense to be defined as some part of pair may be readable on both cpu and gpu
* removed redundant specialization of `for_each`, which was a copy-paste only differing by not calling `local` and using `distributed_range` instead of`distributed_contiguous_range`. This specialization looked to me as a workaround for compilation problems with missing `local()` and moreover `distributed_range` concept seems a mistake to me as code inside required random access in `fn(*(begin + idx))` line
* removed requirement for `distributed_contiguous_range` concept to have `local` call. I see no reason why `distributed_range` does not require `local` while `distributed_contiguous_range` requires. They should differ only by random accessibility and now they do. `local` is not a good requirement as range may be result of `zip_view` or `enumerate`


include/dr/concepts/concepts.hpp
```diff
@@ -36,11 +36,7 @@ concept distributed_iterator =
 
 template <typename R>
 concept remote_contiguous_range =
-    remote_range<R> && rng::random_access_range<R> &&
-    requires(R &r) {
-      lib::ranges::rank(r);
-      { lib::ranges::local(r) } -> rng::contiguous_range;
-    };
+    remote_range<R> && rng::random_access_range<R>;
```
BenBrock, line 39: We'll talk about this tomorrow, but the important part of the distributed contiguous range concept is actually that there's a `local()` CPO that returns a contiguous range, so we don't want to remove that.  The key idea is that the local iterator might be a lot more efficient than the remote iterator.  (This is particularly true for multi-node, but may come up multi-GPU as well.)

We might consider adding a new concept `locally_accessible_range` or something that checks for `local`, since there might be non-contiguous ranges that have more efficient local iterators than remote.

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -13,84 +13,34 @@
 
 namespace shp {
 
-template <typename ExecutionPolicy, lib::distributed_contiguous_range R,
-          typename Fn>
-void for_each(ExecutionPolicy &&policy, R &&r, Fn &&fn) {
+template <lib::distributed_contiguous_range R, typename Fn>
```
BenBrock, line 16: I understand the urge to simplify, but I think we actually do want to leave the signature as `typename ExecutionPolicy`, as we will be adding other execution policies later, and we also want it to resemble the [standard library's interface](https://en.cppreference.com/w/cpp/algorithm/for_each).

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -13,84 +13,34 @@
...
+void for_each(const device_policy &policy, R &&r, Fn &&fn) {
+  std::span<const cl::sycl::device> devices = policy.get_devices();
```
BenBrock, line 18: We haven't defined an execution policy interface, but I'd prefer to leave it a bit more generic, rather than always requiring a `std::span`.  We could use something like

```cpp
rng::forward_range auto devices = policy.get_devices();
```

to be a bit more generic.

include/dr/shp/algorithms/for_each.hpp
lslusarczyk: Restored `ExecutionPolicy` template in order to give nicer compilation errors.
Removed unnecessary `if constexpr` as we discussed in the call.

include/dr/shp/algorithms/for_each.hpp
lslusarczyk: ok, restored `auto`

include/dr/concepts/concepts.hpp
lslusarczyk: Then why does `distributed_range` not require `local()`?

What about `enumerate` or `zip` result? Shall we implement `zip_view` in such a way that the result of `enumerate`/`zip_view` supports `local` call?

We may discuss this on our next call if it is easier not to write long explanations here/

BenBrock: I feel like `local()` should be optional, but not required, for `remote_range`.  We could add another concept called `locally_convertible_range` or something like that.  I haven't added it because it's not the highest priority at the moment.  (The primary advantage of `local()` is for the distributed memory/MPI case.)

`local()` has particular importance for shp at the moment because of a bug in our SYCL runtime that we can only get around by using raw pointers.

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -16,74 +16,23 @@ namespace shp {
...
-      event.wait();
-    }
-  } else {
-    assert(false);
+  for (auto &&segment : lib::ranges::segments(r)) {
+    assert(rng::size(segment) > 0);
+    events.emplace_back(
+        sycl::queue(devices[lib::ranges::rank(segment)])
+            .parallel_for(rng::size(segment), [=](cl::sycl::id<1> idx) {
+              fn(*(rng::begin(segment) + idx));
```
BenBrock, line 30: It's actually important that we extract the iterator and only copy that to the GPU.  The segment itself is likely not device copyable.  It's also important that we call `local()` on the iterator to get a raw pointer, as there's a bug in our SYCL runtime that may keep this from running otherwise.

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -16,74 +16,23 @@ namespace shp {
...
+            }));
   }
-}
-
-template <typename ExecutionPolicy, lib::distributed_range R, typename Fn>
-void for_each(ExecutionPolicy &&policy, R &&r, Fn &&fn) {
```
BenBrock, line 55: We do still need to have two versions of `for_each`, one for distributed contiguous ranges and one for normal distributed ranges.  Let's talk about this in our next chat.

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -16,74 +16,23 @@ namespace shp {
 template <typename ExecutionPolicy, lib::distributed_contiguous_range R,
           typename Fn>
 void for_each(ExecutionPolicy &&policy, R &&r, Fn &&fn) {
-
-  namespace sycl = cl::sycl;
```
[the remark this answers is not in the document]
rscohn2: We should be doing `#include <sycl/sycl.hpp>` and `sycl::` I was planning a mass cleanup. Here is the spec:
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:what-changed-between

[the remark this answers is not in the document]
lslusarczyk: reported [issue](https://github.com/orgs/oneapi-src/projects/15/views/1?pane=issue&itemId=21645945) for this in our backlog

include/dr/shp/algorithms/for_each.hpp
lslusarczyk: as we discussed in person - I've created `get_local_segment` and two versions were written shorter

BenBrock:
Looks good! I made a couple of small changes and will merge this PR now.

PR #48, opened 2023-02-23
add basic unit tests for a distributed vector

haichangsi:
added unit tests for checking the size and the segments number of a distributed vector

### DIFF
--- a/include/dr/shp/distributed_vector.hpp
+++ b/include/dr/shp/distributed_vector.hpp
@@ -187,13 +187,15 @@ public:
   }
 
   iterator end() {
-    return iterator(segments_, size() / segment_size_, size() % segment_size_,
-                    segment_size_);
+    return size_ ? iterator(segments_, size() / segment_size_,
+                            size() % segment_size_, segment_size_)
+                 : begin();
   }
 
   const_iterator end() const {
-    return const_iterator(segments_, size() / segment_size_,
-                          size() % segment_size_, segment_size_);
+    return size_ ? const_iterator(segments_, size() / segment_size_,
+                                  size() % segment_size_, segment_size_)
+                 : begin();
   }
 
 private:
--- a/test/gtest/shp/containers.cpp
+++ b/test/gtest/shp/containers.cpp
@@ -2,38 +2,34 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include "shp-tests.hpp"
+#include "containers.hpp"
 
-using V = std::vector<int>;
-using CV = const std::vector<int>;
-using CVR = const std::vector<int> &;
+TYPED_TEST_SUITE(DistributedVectorTest, AllocatorTypes);
 
-using DV = shp::distributed_vector<int>;
-using CDV = const shp::distributed_vector<int>;
-using CDVR = const shp::distributed_vector<int> &;
-
-TEST(DistributedVector, is_random_access_range) {
-  static_assert(rng::random_access_range<V>);
-  static_assert(rng::random_access_range<CV>);
-  static_assert(rng::random_access_range<CVR>);
-
-  static_assert(rng::random_access_range<DV>);
-  static_assert(rng::random_access_range<CDV>);
-  static_assert(rng::random_access_range<CDVR>);
+TYPED_TEST(DistributedVectorTest, is_random_access_range) {
+  static_assert(rng::random_access_range<typename TestFixture::LocalVec>);
+  static_assert(rng::random_access_range<const typename TestFixture::LocalVec>);
+  static_assert(
+      rng::random_access_range<const typename TestFixture::LocalVec &>);
+  static_assert(rng::random_access_range<typename TestFixture::DistVec>);
+  static_assert(rng::random_access_range<const typename TestFixture::DistVec>);
+  static_assert(
+      rng::random_access_range<const typename TestFixture::DistVec &>);
 }
 
-TEST(DistributedVector, Segments) {
-  DV dv_a(10);
-  std::iota(dv_a.begin(), dv_a.end(), 20);
+TYPED_TEST(DistributedVectorTest,
+           segments_begins_where_its_creating_iterator_points_to) {
+  typename TestFixture::DistVec dv(10);
+  std::iota(dv.begin(), dv.end(), 20);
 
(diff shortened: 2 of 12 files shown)


include/dr/shp/distributed_vector.hpp
```diff
@@ -175,6 +175,8 @@ struct distributed_vector {
 
   size_type size() const noexcept { return size_; }
 
+  size_type segment_size() const noexcept { return segment_size_; }
+
```
rscohn2, line 179: Will this make it possible for the test to access private members instead of exporting public interfaces: https://stackoverflow.com/questions/46107103/accessing-private-members-with-googletest?

lslusarczyk: Let's test public interface only. This method can be removed. Segment size can be checked by calling `segments` and checking its size.

rscohn2:
The earlier clang fail was a problem in the test. Rerunning made it go away. I think this fixes the problem: #49

include/dr/shp/distributed_vector.hpp
BenBrock: I'd prefer not to make the public-facing API too big, since calling `segments().size()` can already do this without introducing overhead.

haichangsi: Done. Thanks. Please re-review this pr.

test/gtest/shp/containers.cpp
```diff
@@ -30,3 +31,49 @@ TEST(ShpTests, DistributedVectorSegments) {
...
+TEST(ShpTests, DistributedVectorAllSegmentsSizeExceptLast) {
+  const int n = 10;
+  DV dv_a(n);
+  auto segments = dv_a.segments();
+  std::size_t expected_segment_size = (n + shp::nprocs() - 1) / shp::nprocs();
+  std::size_t segments_number =
+      n / expected_segment_size + (n % expected_segment_size != 0);
+
+  for (int i = 0; i < segments_number - 1; i++) {
+    EXPECT_TRUE(segments[i].size() == expected_segment_size);
```
rscohn2, line 44:
```suggestion
    EXPECT_EQ(segments[i].size(), expected_segment_size);
```

test/gtest/shp/containers.cpp
```diff
@@ -30,3 +31,49 @@ TEST(ShpTests, DistributedVectorSegments) {
...
+
+TEST(ShpTests, DistributedVectorSegmentsNumber) {
+  const int n = 10;
+  DV dv_a(n);
+  auto segments = dv_a.segments();
+  auto segments_number = segments.size();
+  std::size_t expected_segment_size = (n + shp::nprocs() - 1) / shp::nprocs();
+  std::size_t expected_segments_number =
+      n / expected_segment_size + (n % expected_segment_size != 0);
+  EXPECT_TRUE(segments_number == expected_segments_number);
```
rscohn2, line 56:
```suggestion
  EXPECT_EQ(segments_number, expected_segments_number);
```

test/gtest/shp/containers.cpp
```diff
@@ -30,3 +31,49 @@ TEST(ShpTests, DistributedVectorSegments) {
...
+}
+
+TEST(ShpTests, DistributedVectorLastSegmentSize) {
+  const size_t n = 10;
+  DV dv_a(n);
+  auto segments = dv_a.segments();
+  std::size_t segment_size = (n + shp::nprocs() - 1) / shp::nprocs();
+  std::size_t segments_number = n / segment_size + (n % segment_size != 0);
+  std::size_t expected_last_element_size = n % segment_size;
+  EXPECT_TRUE(segments[segments_number - 1].size() ==
```
rscohn2, line 77: Use EXPECT_EQ

test/gtest/shp/containers.cpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 #include "shp-tests.hpp"
+#include <typeinfo>
```
rscohn2, line 6: Why do you need this?

rscohn2 (commented):
The fail looks like a problem in the logic of the test, otherwise ok.

test/gtest/shp/containers.cpp
haichangsi: I forgot to delete this, thanks, done

BenBrock:
The tests `DistributedVectorLastSegmentSize`, `DistributedVectorSegmentsNumber`, and `DistributedVectorLastSegmentSize` actually mandate that the segments be laid out in a particular fashion.  (That each segment has a particular size.)  I don't think we want to do that, since different implementations might potentially make the segments larger or smaller based on how much the vector is expected to grow.

Maybe you could replace these with a test that checks that all the segments together are the right size?

rscohn2:
> Maybe you could replace these with a test that checks that all the segments together are the right size?

This test checks size and contents: https://github.com/oneapi-src/distributed-ranges/blob/568dc5bd8cd90bef8b5bc717faf5431c6df013f0/test/gtest/include/common-tests.hpp#L41
We should be doing that for all view tests. You could add it here: https://github.com/oneapi-src/distributed-ranges/blob/main/test/gtest/include/common/view-tests.hpp and in dv-tests.hpp

It would be good to have some tests that the data is split up, even if the API does not require it. How about number of segments == number of ranks and all segments have a size > 0. The same logic could go in `check_segments`

lslusarczyk:
> The tests `DistributedVectorLastSegmentSize`, `DistributedVectorSegmentsNumber`, and `DistributedVectorLastSegmentSize` actually mandate that the segments be laid out in a particular fashion. (That each segment has a particular size.) I don't think we want to do that, since different implementations might potentially make the segments larger or smaller based on how much the vector is expected to grow.

Unittest of DistributedVector checks this one exact implementation. It is a self-validating documentation how class is implemented. Next implementations will get its own unittests testing if segments are laid in these next implementations fashions. We can also have extra tests checking requirements of all implementations and execute same tests for two implementations. But now there is only one, so I don't separate it.

> Maybe you could replace these with a test that checks that all the segments together are the right size?

Thanks for the idea. I've added cases like this. As Robert advised - I used `check_segments`. ... and one more bug found due to this :)

rscohn2 (approved):

PR #50, opened 2023-02-23
reimplement mhp distributed vector

rscohn2:
Directly implements segments instead of using a chunk view on the distributed iterator
uses `normal_distributed_iterator` to provide the distributed iterator

Uses free function for `fence()` that operates on all fences. same for `barrier()`


include/dr/details/normal_distributed_iterator.hpp
```diff
@@ -62,6 +60,7 @@ class normal_distributed_iterator_accessor {
           std::min(-offset, difference_type(idx_) + 1);
```
rscohn2, line 60: @BenBrock The `normal_distributed_iterator` assumes that it can join the individual segments into an iterator space for the distributed data structure, and the segments can be any length. This would not work well for interleave with a block size of 1. We would need $n$ segments to describe it. For cyclic, there should only be 1 segment per rank, and we need a `cyclic_distributed_iterator` that knows how to map the distributed iterator space to a segment and offset.

rscohn2: A related issue is that MHP can't handle a zip where a segment contains data in multiple ranks. If I know that the segments are not single rank, then I can use the fallback--direct use of the distributed iterator/range. Is there some way to mark this so I do not have to check the segments at runtime? Currently, I use an empty segments range as a marker.

test/gtest/mhp/views.cpp
```diff
@@ -99,8 +101,7 @@ TEST(MhpTests, Drop) {
   }
 }
 
-#if 0
-TEST(MhpTests, Transform) {
+TEST(MhpTests, TransformView) {
```
rscohn2, line 104: @BenBrock: MHP can use your transform view now.

PR #52, opened 2023-02-24
remove redundant includes

rscohn2:
all shp programs include shp.hpp, so delete the includes that are in shp.hpp and in other files. Lately, I have been moving includes from shp to details, and this will make it easier. Are you OK with this @BenBrock ?


BenBrock:
I would prefer not to do it this way.  Could we maybe set up more of a hierarchy instead, like adding a `shp/views.hpp` that includes all the shp views, so you still only have to change one line, but we still keep the header files self-contained?

Personally I think headers should be self-contained and able to be included individually.  If the headers only work when included in a particular order, it can get pretty confusing, and since clang-format is already re-ordering our includes, I think that could turn into a headache.

rscohn2:
Let's talk about it next Tuesday.

> clang-format is already re-ordering our includes

Here is how you stop the sorting: https://github.com/oneapi-src/distributed-ranges/blob/e2948567e6b09e392d9618ee5684cca0c58bf8ab/include/dr/shp/shp.hpp#L7

clang-format does not sort across blank lines, which usually separates system includes and internal includes.

PR #54, opened 2023-02-25
turn on Wall/Werror for shp and fix issues

rscohn2:
Turned on warnings for shp-specific code and fixed issues. You can disable specific warnings this way: https://github.com/oneapi-src/distributed-ranges/blob/8c9852e73518107aaa88d959eef0d2be09fdf53e/CMakeLists.txt#L143

None of the issues were bugs, but they are still worth fixing.

* unused variables
* reference to uninitialized variable
* initializer list order does not match constructor order

```
/nfs/site/home/rscohn1/local/projects/dds/dr/include/./dr/shp/device_span.hpp:56:14: error: field 'rank_' will be initialized afte      base 'shp::span<int, device_ptr<int>>' [-Werror,-Wreorder-ctor]
    :        rank_(lib::ranges::rank(r)),
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
             shp::span<T, Iter>(rng::begin(r), rng::size(r))
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,12 +127,6 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 
   include(GoogleTest)
 
-  if(ENABLE_SYCL)
-    add_subdirectory(examples/shp)
-    add_subdirectory(test/gtest/shp)
-  endif()
-
-  # shp has warnings, but do for everything else
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wall>)
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Werror>)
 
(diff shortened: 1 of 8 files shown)


rscohn2:
@BenBrock SHP-3 sometimes gets errors in inclusive scan: https://github.com/oneapi-src/distributed-ranges/actions/runs/4270293192/jobs/7434084747
Github hosted runners only have 2 cores and shp3 uses 3 devices so there might be some timing issues.

BenBrock:
Weird, I haven't gotten this error, but will try to debug locally.  Do we have a similar system to this at Intel?

rscohn2:
I wouldn't spend time trying to reproduce it. Getting a machine with fewer cores might help. I tried doing:
```
taskset -c 0-3 ./shp-tests 3
```
It fails in `create_subdevice`. That looks like a different issue, and possibly an issue with sycl runtime, not application code.

PR #60, opened 2023-02-27
Investigate failing copy test

BenBrock:
Some of the new copy tests are failing when run on multiple GPUs.  I'm trying to investigate why.

### DIFF
--- a/include/dr/shp/copy.hpp
+++ b/include/dr/shp/copy.hpp
@@ -11,9 +11,6 @@
 #include <memory>
 #include <type_traits>
 
-#include <fmt/core.h>
-#include <fmt/ranges.h>
-
 namespace shp {
 namespace __detail {
 
--- a/include/dr/shp/copy.hpp
+++ b/include/dr/shp/copy.hpp
@@ -11,6 +11,9 @@
 #include <memory>
 #include <type_traits>
 
+#include <fmt/core.h>
+#include <fmt/ranges.h>
+
 namespace shp {
 namespace __detail {
 
@@ -75,19 +78,24 @@ template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
   requires __detail::is_syclmemcopyable<std::iter_value_t<InputIt>,
                                         std::iter_value_t<OutputIt>>
 cl::sycl::event copy_async(InputIt first, InputIt last, OutputIt d_first) {
-  auto segments = lib::ranges::segments(d_first);
-  auto segment = std::begin(segments);
+  auto &&segments = lib::ranges::segments(d_first);
+  auto segment_iter = rng::begin(segments);
 
   std::vector<cl::sycl::event> events;
 
   while (first != last) {
-    const std::size_t n_to_copy =
-        std::min<size_t>((*segment).size(), last - first);
+    auto &&segment = *segment_iter;
+    auto size = std::distance(rng::begin(segment), rng::end(segment));
 
-    events.push_back(
-        shp::copy_async(first, first + n_to_copy, (*segment).begin()));
+    std::size_t n_to_copy = std::min<size_t>(size, std::distance(first, last));
 
-    ++segment;
+    auto local_last = first;
+    std::advance(local_last, n_to_copy);
+
+    events.emplace_back(
+        shp::copy_async(first, local_last, rng::begin(segment)));
+
+    ++segment_iter;
     std::advance(first, n_to_copy);
   }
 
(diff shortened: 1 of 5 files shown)


rscohn2:
I updated dpl. Could that be the problem

BenBrock:
> I updated dpl. Could that be the problem

I think it's unrelated to oneDPL, as we simply weren't testing this case before (copying from local->dist with USM shared).  I'm not sure why it's failing, but it looks to me like a SYCL issue.  Either the way we're calling SYCL memcpy is wrong, or there's some issue with USM shared when you copy to more than one GPU.

BenBrock:
I did a bit more investigation and ended up [filing a bug](https://jira.devtools.intel.com/projects/CMPLRLLVM/issues/CMPLRLLVM-45277?filter=allissues) with the SYCL runtime folks.  The issue is related to SYCL USM shared memory not behaving the way we expect, and not anything we did.

All the changes to copy were fine and unrelated.

In the meantime I think there's not much to do, just hope there's a fix at some point.  I do think we should switch to USM device as the default for everything `shp`, as this is just one of multiple problems I've encountered with USM shared.

lslusarczyk:
I am still concerned about using default constructor when creating queue. This is what @BenBrock and I were talking last week. How can sycl know which device pointer is residing? I think it is the reason of the bug.

I asked about that in JIRA ticker. We may fix queue constructor and check if problem will disappear.

include/dr/shp/copy.hpp
```diff
@@ -75,19 +75,24 @@ template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
   requires __detail::is_syclmemcopyable<std::iter_value_t<InputIt>,
                                         std::iter_value_t<OutputIt>>
 cl::sycl::event copy_async(InputIt first, InputIt last, OutputIt d_first) {
-  auto segments = lib::ranges::segments(d_first);
-  auto segment = std::begin(segments);
+  auto &&segments = lib::ranges::segments(d_first);
```
lslusarczyk, line 78: `auto &&` is for forwarding only (+ some exceptions with good reasons) and should not be used for local variables. See [Why should one never use auto&& for local variables?](https://stackoverflow.com/questions/26355406/why-should-one-never-use-auto-for-local-variables).

include/dr/shp/copy.hpp
```diff
@@ -75,19 +75,24 @@ template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
...
+  auto segment_iter = rng::begin(segments);
```
lslusarczyk, line 79: Could you please explain the difference and why to use one over the other?

include/dr/shp/copy.hpp
```diff
@@ -75,19 +75,24 @@ template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
...
 
   while (first != last) {
-    const std::size_t n_to_copy =
-        std::min<size_t>((*segment).size(), last - first);
+    auto &&segment = *segment_iter;
+    auto size = std::distance(rng::begin(segment), rng::end(segment));
 
-    events.push_back(
-        shp::copy_async(first, first + n_to_copy, (*segment).begin()));
+    std::size_t n_to_copy = std::min<size_t>(size, std::distance(first, last));
```
lslusarczyk, line 87: `size` is a variable used in one place only. If we do not need its name for improve readability of the code I suggest putting `std::distance(rng::begin(segment), rng::end(segment))` inside `min`.
... or making `std::distance(first, last)` also a variable to be consistent (havew something like `min(segment_size, remainder_size`).

But this is just cosmetic. I'm ok with this also .

include/dr/shp/copy.hpp
```diff
@@ -75,19 +75,24 @@ template <std::forward_iterator InputIt, lib::distributed_iterator OutputIt>
...
   std::vector<cl::sycl::event> events;
 
   while (first != last) {
-    const std::size_t n_to_copy =
-        std::min<size_t>((*segment).size(), last - first);
+    auto &&segment = *segment_iter;
+    auto size = std::distance(rng::begin(segment), rng::end(segment));
 
-    events.push_back(
-        shp::copy_async(first, first + n_to_copy, (*segment).begin()));
```
lslusarczyk, line 88: Are we not allowed to use `iter + size`? If we can't - why? If we can - concise code is more readable to me.

include/dr/shp/copy.hpp
lslusarczyk: I see an [article](https://stackoverflow.com/questions/62183286/what-is-the-difference-between-stdrangesbegin-and-stdbegin) about it. I'll read.

include/dr/shp/copy.hpp
BenBrock: `segments()` returns a viewable range, which may include an lvalue reference.  I use `auto&&` here because it will bind `auto` to an lvalue reference, if one is returned, whereas `auto` will always bind to the value type.

include/dr/shp/copy.hpp
BenBrock: I would like the distributed copy functions to work for forward iterators, so I want to avoid using expressions like `iter + size` that only work on random access iterators.  `std::advance` will do the same thing, but work for all iterator types.

include/dr/shp/copy.hpp
BenBrock: Segments are not necessarily sized, so they might not have `rng::size` or `.size()`.  `std::distance()` will work for any forward range.

At some point we really want our own wrapper that will return `rng::size()` for a sized range and `std::distance()` otherwise.

include/dr/shp/copy.hpp
```diff
@@ -109,17 +114,20 @@ template <lib::distributed_iterator InputIt, std::forward_iterator OutputIt>
   requires __detail::is_syclmemcopyable<std::iter_value_t<InputIt>,
                                         std::iter_value_t<OutputIt>>
 cl::sycl::event copy_async(InputIt first, InputIt last, OutputIt d_first) {
+  auto dist = std::distance(first, last);
```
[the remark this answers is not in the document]
BenBrock: Same as above---I want to keep the distributed algorithms working with any forward range, not just random access ranges.  `std::distance()` works with any forward range, while `last-first` requires a random access range.

PR #67, opened 2023-03-01
add support for common tests

rscohn2:
Support for sharing tests between shp, mhp CPU, mhp SYCL

Using gtest support for type parameters. Generic tests are defined in `test/gtest/include/common`. They are instantiated with types in `test/gtest/mhp/mhp-tests.cpp` and `test/gtest/shp/shp-tests.cpp`

Because of differences in namespaces and execution policies, we need to wrap algorithms to allow generic tests.


test/gtest/mhp/mhp-tests.cpp
```diff
@@ -4,6 +4,38 @@
...
+template <typename T>
+struct CommonTestConfigSYCL : public CommonTestConfigBase<T> {
+  using DV = mhp::distributed_vector<T, mhp::sycl_shared_allocator<T>>;
+  using DVA = DV;
+  static auto policy() { return mhp::device_policy(); }
+};
+#endif
+using Common_Types = ::testing::Types<
+#ifdef SYCL_LANGUAGE_VERSION
+    CommonTestConfigSYCL<int>, CommonTestConfigSYCL<float>,
```
lslusarczyk, line 34: There is no special handling in the code for either 'int' or 'float' - the same code is executed. Code or branch coverage will be the same if just one is used, therefore there is no point in having `CommonTestConfigSYCL<int>` and `CommonTestConfigSYCL<float>`, but since we have to use some type with `CommonTestConfigSYCL` and `CommonTestConfigCPU` it is a good idea to use one with one type and another with another (best also having differents size).

test/gtest/include/common/dv-tests.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+  static_assert(rng::random_access_range<decltype(dv.segments())>);
+  static_assert(rng::random_access_range<decltype(dv.segments()[0])>);
+  static_assert(rng::viewable_range<decltype(dv.segments())>);
+  static_assert(std::forward_iterator<DVI>);
+  static_assert(rng::forward_range<DV>);
+  static_assert(rng::random_access_range<DV>);
+
+  static_assert(lib::distributed_iterator<decltype(dv.begin())>);
+  // static_assert(lib::remote_iterator<decltype(dv.segments()[0].begin())>);
```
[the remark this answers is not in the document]
rscohn2: ok

test/gtest/include/mhp-tests.hpp
```diff
@@ -2,14 +2,18 @@
...
-#include "common-tests.hpp"
-
 extern MPI_Comm comm;
 extern int comm_rank;
 extern int comm_size;
+
+#include "common-tests.hpp"
+
+// MHP specific tests
+template <typename T> class MhpTests : public testing::Test {
```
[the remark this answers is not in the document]
rscohn2: ok

test/gtest/mhp/mhp-tests.cpp
rscohn2: I don't think there is a lot of value, but there is some value in testing different types.

test/gtest/mhp/mhp-tests.cpp
```diff
@@ -4,6 +4,38 @@
 
 #include "mhp-tests.hpp"
 
+// Instantiate MHP-specific configurations for common tests
+template <typename T> struct CommonTestConfigBase {
+  using DV = mhp::distributed_vector<T>;
```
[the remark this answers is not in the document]
rscohn2: Having a separate test suite for allocators would make this easier.

test/gtest/include/common/dv-tests.hpp
```diff
@@ -0,0 +1,30 @@
...
+  static_assert(lib::distributed_contiguous_range<DV>);
+}
+
+TYPED_TEST_P(CommonTests, DistributedVectorConstructors) {
```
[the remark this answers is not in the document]
rscohn2: tests are not complete. Assertions are checks, even if they are compile-time.

test/gtest/mhp/mhp-tests.cpp
lslusarczyk: It is enough to test vector of one particular size in UT unless it has some special handling depending on size.
For the same reason I also do not test templates with two types if handling them is identical.

But I don't insist - it's cosmetic.

rscohn2: The fuzzing varies the size and alignment of subranges. It finds bugs, because distribution slices the ranges into segments and there are boundary conditions to get right. It would be good to write unit tests that accept parameters to get better coverage. I don't think gtest parameters will work for fuzzing, maybe global variables.

lslusarczyk: My intuition is that fuzz tests are something different than unittests (unittests should be repeatable, non-random, simple, extremely fast). For quick search I see that google has separate framework for it https://github.com/google/fuzztest and there are some docs there about fuzz testing approaches, best practises, which I need to read about.

rscohn2:
google fuzztest looks good. It is a good match for the way we did fuzzing to satisfy SDL.

> Fuzz tests are parameterized unit tests, also called property-based tests. 
https://github.com/google/fuzztest/blob/main/doc/fuzz-test-macro.md

Not all tests need to be fuzzable, but we should fuzz test all the algorithms. It's another reason not to use initializer lists for inputs/results.

test/gtest/mhp/mhp-tests.cpp
rscohn2: google fuzztest looks good. It is a good match for the way we did fuzzing to satisfy SDL.

Fuzz tests are parameterized unit tests, also called property-based tests.
https://github.com/google/fuzztest/blob/main/doc/fuzz-test-macro.md

Not all tests need to be fuzzable, but we should fuzz test all the algorithms. It's another reason not to use initializer lists for inputs/results.

PR #73, opened 2023-03-02
update DPL, cleanup deprecation warnings

rscohn2:
@BenBrock: I am updating DPL to latest and adding a patch for a warning.

Removing the remaining use of sycl 2020 deprecations.


rscohn2:
inclusive scan did not fail with this PR. It might be fixed in the latest DPL, or maybe it was luck.

PR #78, opened 2023-03-03
common test subrange, take, drop

rscohn2:
take is broken in MHP and disabled

### DIFF
--- a/test/gtest/include/common-tests.hpp
+++ b/test/gtest/include/common-tests.hpp
@@ -101,5 +101,7 @@ TYPED_TEST_SUITE_P(CommonTests);
 
 #include "common/algorithm-tests.hpp"
 #include "common/dv-tests.hpp"
-REGISTER_TYPED_TEST_SUITE_P(CommonTests, DistributedVectorRequirements,
-                            DistributedVectorConstructors, ForEach);
+#include "common/view-tests.hpp"
+REGISTER_TYPED_TEST_SUITE_P(CommonTests, DistributedVectorConstructors,
+                            DistributedVectorRequirements, Drop, ForEach,
+                            Subrange, DISABLED_Take);
--- a/test/gtest/include/common/view-tests.hpp
+++ b/test/gtest/include/common/view-tests.hpp
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+TYPED_TEST_P(CommonTests, Subrange) {
+  using DV = typename TypeParam::DV;
+  using V = typename TypeParam::V;
+
+  std::size_t n = 10;
+
+  auto neg = [](auto &v) { v = -v; };
+  DV dv_a(n);
+  TypeParam::iota(dv_a, 100);
+  TypeParam::for_each(TypeParam::policy(),
+                      rng::subrange(dv_a.begin() + 1, dv_a.end() - 1), neg);
+
+  if (comm_rank == 0) {
+    V a(n), a_in(n);
+    rng::iota(a, 100);
+    rng::iota(a_in, 100);
+    rng::for_each(rng::subrange(a.begin() + 1, a.end() - 1), neg);
+    EXPECT_TRUE(unary_check(a_in, a, dv_a));
+  }
+}
+
+TYPED_TEST_P(CommonTests, Drop) {
+  using DV = typename TypeParam::DV;
+  using V = typename TypeParam::V;
+
+  std::size_t n = 10;
+
+  auto neg = [](auto &v) { v = -v; };
+  DV dv_a(n);
+  TypeParam::iota(dv_a, 100);
+  TypeParam::for_each(TypeParam::policy(), rng::views::drop(dv_a, 2), neg);
+
+  if (comm_rank == 0) {
+    V a(n), a_in(n);
+    rng::iota(a, 100);
+    rng::iota(a_in, 100);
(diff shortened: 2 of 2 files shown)


test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
...
+  DV dv_a(n);
+  TypeParam::iota(dv_a, 100);
+  TypeParam::for_each(TypeParam::policy(),
+                      rng::subrange(dv_a.begin() + 1, dv_a.end() - 1), neg);
+
+  if (comm_rank == 0) {
+    V a(n), a_in(n);
+    rng::iota(a, 100);
+    rng::iota(a_in, 100);
+    rng::for_each(rng::subrange(a.begin() + 1, a.end() - 1), neg);
```
lslusarczyk, line 21: We do not test `for_each`. Let's remove its usage from `subrange` tests. See [unittests of subrange](https://github.com/ericniebler/range-v3/blob/master/test/view/subrange.cpp) for examples how to write simple checks on subrange results (check begin/end, dereference).

test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
...
+    EXPECT_TRUE(unary_check(a_in, a, dv_a));
```
lslusarczyk, line 22: Check look perfectly readable when having exact values which we expect, like
```
EXPECT_TRUE(equal(dist_vec, V{2, 3, 4}));
```
... so one has no to trace what happened to to a_in, a, dv_a during the test and what does `unary_check` do.
See [slice unittest](https://github.com/ericniebler/range-v3/blob/master/test/view/slice.cpp) for a good example on how checks on expected results can look like.

test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
...
+  auto neg = [](auto &v) { v = -v; };
+  DV dv_a(n);
+  TypeParam::iota(dv_a, 100);
+  TypeParam::for_each(TypeParam::policy(), rng::views::drop(dv_a, 2), neg);
+
+  if (comm_rank == 0) {
+    V a(n), a_in(n);
+    rng::iota(a, 100);
+    rng::iota(a_in, 100);
+    rng::for_each(rng::views::drop(a, 2), neg);
```
lslusarczyk, line 41: Same comments as for `subrange`. Ideally there should be no `take`, `iota`, `foreach` here, just `drop` and simple begin/end/dereference/equality checks

test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
...
+    EXPECT_TRUE(unary_check(a_in, a, dv_a));
+  }
+}
+
+TYPED_TEST_P(CommonTests, DISABLED_Take) {
```
lslusarczyk, line 46: Why is this test commented out? Could we establish a rule, that we comment out a test always with a comment saying about the reason?

test/gtest/include/common-tests.hpp
```diff
@@ -101,5 +101,7 @@ TYPED_TEST_SUITE_P(CommonTests);
 
 #include "common/algorithm-tests.hpp"
 #include "common/dv-tests.hpp"
-REGISTER_TYPED_TEST_SUITE_P(CommonTests, DistributedVectorRequirements,
-                            DistributedVectorConstructors, ForEach);
+#include "common/view-tests.hpp"
+REGISTER_TYPED_TEST_SUITE_P(CommonTests, DistributedVectorConstructors,
```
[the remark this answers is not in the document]
rscohn2: ok

test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
+// SPDX-FileCopyrightText: Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+TYPED_TEST_P(CommonTests, Subrange) {
+  using DV = typename TypeParam::DV;
```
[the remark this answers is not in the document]
rscohn2: I agree in general. In this case, it is the `typename` that was the motivation for separate definition.

test/gtest/include/common/view-tests.hpp
```diff
@@ -0,0 +1,64 @@
...
+  using V = typename TypeParam::V;
+
+  std::size_t n = 10;
+
+  auto neg = [](auto &v) { v = -v; };
+  DV dv_a(n);
+  TypeParam::iota(dv_a, 100);
```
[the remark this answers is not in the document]
rscohn2: Using initializer lists everywhere would be inflexible and tedious. We need a standard way to initialize. I don't like fill because it does not catch position errors.

test/gtest/include/common/view-tests.hpp
rscohn2: check_equal and unary_check are similar, I don't see a difference.

lslusarczyk: `equal` is standard and has obvious semantic, reader has to see `unary_check` for implementation to see that 1st param is just for printing it out and 2nd, 3rd are for equality check

rscohn2: I don't object if you want to port or adapt the range v3 tests or write tests in that style. I started writing the DR tests that way and found they weren't what I needed to find and fix the bugs. There are complex interactions between views and algorithms, off by 1 and other boundary conditions and I need to write tests that are flexible and combine views and algorithms to find the issues.

PR #79, opened 2023-03-03
Implement Sparse Matrix IO

BenBrock:


CMakeLists.txt
```diff
@@ -125,6 +125,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wall>)
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Werror>)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-unused-local-typedef>)
```
BenBrock, line 128: @rscohn2 Is this the correct way to ignore an error?  Or should I be adding this flag to a smaller subset?

lslusarczyk: let's please disable compiler warnings in as limited scope as possible - a file or maybe just `[[maybe_unused]]` in the code will do the work? Or add a fake usage in the code to silence warning?

rscohn2: Occasionally, something is unused because of a logic error or typo so I would not want to turn it off everywhere. Some possibilites:
* put it here: https://github.com/oneapi-src/distributed-ranges/blob/ffcfb71e81eaf04e3b62a07079bf2b6d10adc720/CMakeLists.txt#L142 and move the clause to the end of the file so it would not affect mhp
* pragma for limited scope: https://stackoverflow.com/questions/3378560/how-to-disable-gcc-warnings-for-a-few-lines-of-code but I don't know if it works for clang.  
* make it a templated type alias with a dummy argument?

[maybe_unused] seems like the best choice.

BenBrock: Sounds good---I will just add `[maybe_unused]` for now.  (Did not even know that existed!)

PR #80, opened 2023-03-03
Updates to use `rng::distance` and `rng::advance`

BenBrock:


rscohn2:
How could we test for improperly assuming random access iterators?

BenBrock:
One option might be to write tests that pass in forward iterators.  Probably not the highest priority, since I'd expect poor performance with things that aren't at least random access.

lslusarczyk:
from discussion: filter_view, drop_while, take_while will produce non-random access iters

PR #96, opened 2023-03-07
test restructuring

rscohn2:
Changes in response to review comments.

Break tests into separate suites for each algorithm/view. Most suites have a single test right now. But that should grow over time. This structure works out better because the suites are more self-contained and in separate files. When adding a new suite., you must include it in `common-tests.hpp` and instantiate it in `mhp-tests.cpp` and `shp-tests.cpp`


lslusarczyk:
Thank you for applying comments. I'll give my feedback tomorrow morning.

lslusarczyk:
I proposed some changes to your PR in [this pull request to this private branch](https://github.com/rscohn2/distributed-ranges/pull/1/files).
- iota moved to be standalone function is shp-tests, IMO shp-tests.hpp and mhp-tests.hpp are better places to keep xhp specific specializations than type customizing the test
- get rid of DV/DVA typedefs - sometimes they are same and thus redundant, sometimes they were different and thus should be different cases, now they test separately when different
- created default_policy helper in tests, as policy is deducible from allocator type. This allowed to pass just simple vector types to INSTANTIATE_TYPED_TESTS_P instead of configs

Please check if it looks sensible to you.

rscohn2:
Closing, #98 has the changes

PR #97, opened 2023-03-07
enable remote iterator check

rscohn2:


rscohn2:
@BenBrock: See the static_assert failt at: https://github.com/oneapi-src/distributed-ranges/actions/runs/4354986221/jobs/7610999857#step:4:346
This is for SHP. I thought that segments()[0].begin() is a remote iterator, but it is a `int *`.  Which iterator should satisfy `remote_iterator`?
This is the test: https://github.com/oneapi-src/distributed-ranges/pull/97/files

BenBrock:
A remote range's iterator doesn't necessarily have to be a remote iterator, and in shp it's usually not a remote iterator.  This allows easily creating remote versions of views and data structures that only add a rank to indicate locality, without modifying the iterator type.

For example, [`device_vector`](https://github.com/oneapi-src/distributed-ranges/blob/main/include/dr/shp/device_vector.hpp) just adds a `rank` method to `vector`.  If we also had to make the iterator a remote iterator, this would be a lot more complicated.

However if the iterator type of some range is a remote iterator, that range will also automatically be a remote range.

BenBrock:
Just saw you asked which iterator should satisfy remote_iterator.  I'm actually not sure if there's one in `shp`---I will look around.  I think most of the iterators in `mhp` should be remote iterators.

rscohn2:
My question should have been: 
Which iterator must satisfy `remote_iterator`?
And the answer and the answer is none. I will delete the check.

rscohn2:
Making the change in #96

PR #98, opened 2023-03-07
mhp reduce

rscohn2:
Implements:
* MHP versions of `std::reduce`
* range versions to match SHP
* *root* versions for MHP (without a root, all ranks compute reduction)
* Tests for all API for SHP/MHP API
 
This should make `reduce` complete for DR.

@BenBrock: Please look over the files that have reduce in the name. The other files are restructuring of tests.

Depends on #96


rscohn2:
mhp sycl is not implemented, but tests pass because it uses usm shared memory. Will work on sycl version.

rscohn2:
Completes #85

PR #101, opened 2023-03-07
enable inclusive scan tests

rscohn2:
Appears to be working with latest DPL


rscohn2:
It passed several times before, but there is still an issue. We can continue to defer tracking it down until it becomes easier to reproduce.

BenBrock:
The random crashes in the build system are very weird, as I have been unable to reproduce them locally.  I will try a few machines when I have time and see if I can reproduce.

PR #105, opened 2023-03-10
2d stencil using 1d arrays

rscohn2:
@mateuszpn: This does not work because of an interaction of zip and subrange, but it does look like it will work for multiple processes.


rscohn2:
Works now. The example found 2 bugs in general code.

PR #108, opened 2023-03-11
check for .begin, .end, .size in include directory

rscohn2:
@BenBrock: I extended the checker to look for `.begin()`, `.end()`, `.size()` in the include directory.  It suggests using `rng::begin`, `rng::end`, `rng::distance`. You can also allow by adding `// dr-style ignore` to the line. For now, I only turned it on in mhp.

The only surprise was having to enable borrowed range for `remote_subrange`, but I think that is correct. You can see that below.

There are cases when `.begin()` is correct, but I think it is better to default to `rng::begin` in our code. Does this look ok? I intend to do the rest of the include directory, but not examples/tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,9 +52,15 @@ repos:
 
 - repo: local
   hooks:
-  - id: dr-style
-    name: dr-style
+  - id: dr-style-all
+    name: dr-style-all
     entry: python scripts/dr-style.py --Werror include test examples
+    language: system
+    pass_filenames: false
+    always_run: true
+  - id: dr-style-include
+    name: dr-style-include
+    entry: python scripts/dr-style.py --Werror --include include/dr/mhp
     language: system
     pass_filenames: false
     always_run: true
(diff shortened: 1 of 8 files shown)


BenBrock:
Sorry, forgot to respond to this.  This looks good to me, so you can go ahead and merge as-is if you like.

How can I enable it for `shp`?  I'd like to do that as well, but will probably need to make some changes.

.pre-commit-config.yaml
```diff
@@ -52,12 +52,18 @@ repos:
...
-    name: dr-style
+  - id: dr-style-all
+    name: dr-style-all
     entry: python scripts/dr-style.py --Werror include test examples
     language: system
     pass_filenames: false
     always_run: true
+  - id: dr-style-include
+    name: dr-style-include
+    entry: python scripts/dr-style.py --Werror --include include/dr/mhp
```
rscohn2, line 63: @BenBrock: This is how pre-commit runs the scan. If you change the path above to /include/dr, it would scan shp as well. Then you could rely on the pre-commit testing or you could execute the above command manually.

BenBrock: Thanks, I will give this a try today.

PR #119, opened 2023-03-14
transform in shp

lslusarczyk:


include/dr/shp/algorithms/transform.hpp
```diff
@@ -0,0 +1,109 @@
...
+namespace shp {
+
+namespace __detail {
+template <class T> T *get_local_pointer(shp::device_ptr<T> r) {
+  return r.local();
+}
+template <class T> T *get_local_pointer(T *t) { return t; }
+} // namespace __detail
+
+void transform(auto &&policy, lib::distributed_range auto &&in,
```
rscohn2, line 19: In an earlier discussion, @BenBrock said that some things could be implemented with `zip` and `for_each`, and we would not have to partition the work. Is that possible for `transform`?

include/dr/shp/distributed_vector.hpp
```diff
@@ -158,7 +158,14 @@ struct distributed_vector {
 
   distributed_vector(std::initializer_list<T> init)
       : distributed_vector(init.size()) {
-    shp::copy(rng::begin(init), rng::end(init), begin());
+    // shp::copy(rng::begin(init), rng::end(init), begin());
```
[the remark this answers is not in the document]
lslusarczyk: Bug here is that copy does not work because of issue described in [this comment](https://github.com/oneapi-src/distributed-ranges/pull/60#issuecomment-1447527300)

Another issue I am facing is that sometimes one test (part_not_aligned) adds 9 instead of 10 into one field when running on a PVC.
```
[ RUN      ] TransformTest/1.part_not_aligned
/nfs/site/home/lslusarc/src/distributed-ranges/test/gtest/shp/transform.cpp:64: Failure
Value of: equal(b, typename TestFixture::LocalVec{9, 9, 9, 9, 9, 11, 12, 13, 9})
  Actual: false (
    [9, 9, 9, 9, 9, 10, 12, 13, 9]
    [9, 9, 9, 9, 9, 11, 12, 13, 9]
  )
Expected: true
[  FAILED  ] TransformTest/1.part_not_aligned, where TypeParam = sycl::_V1::usm_allocator<unsigned long long, (sycl::_V1::usm::alloc)2, 0ul> (18 ms)
```
I shall report these issues as bugs tomorrow. + exclusive scan problem

include/dr/shp/algorithms/transform.hpp
lslusarczyk: good point, we should reuse code if possible - I will try it tomorrow

BenBrock: I had been thinking of something like

```cpp
template <typename R, typename O, typename Fn>
void transform(R&& input_range, O&& output_range, Fn fn) {
  auto zipped_range = shp::views::zip(input_range, output_range);
  shp::for_each(shp::par_unseq, [](auto&& a, auto&& b) { b = fn(a); });
}
```

For a default implementation.  It is true you might want to do something more complicated when they're not aligned, but I think using `zip` makes it a lot easier, since it picks out the overlapping segments for you.

BenBrock: You could then do something like

```cpp
template <typename R, typename O, typename Fn>
void transform(R&& input_range, O&& output_range, Fn fn) {
  auto r = shp::views::zip(input_range, output_range);

  for (auto&& segs : r.zipped_segments()) {
    if (segs[0].rank() == segs[1].rank()) {
      // Both segments are on same rank, just run as normal.
      for_each(...);
    } else {
      // Create local buffer on GPU
      shp::device_vector<T, ...> buffer(segs[0].size(), ...);

      // Run locally as normal
      auto e = for_each_async(...);

      // Copy result from buffer to remote
      shp::copy_async(buffer.begin(), buffer.end(), segs[1].begin(), e);
    }
  }
}
```

rscohn2: > I had been thinking of something like

Yes, that would be much easier to maintain. As you show, it has a default that will give good performance for the typical case (everything aligned) and still allows special case optimizations. There are many variants of transform (https://en.cppreference.com/w/cpp/algorithm/transform) so we need to funnel all the implementations through some common implementation anyway.

test/gtest/shp/transform.cpp
```diff
@@ -0,0 +1,65 @@
+// SPDX-FileCopyrightText: Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+#include "shp-tests.hpp"
+
+namespace drtesting {
+std::optional<bool> expect_fast_transform_path;
+}
+
+#include "dr/shp/algorithms/transform.hpp"
```
rscohn2, line 10:
```suggestion
```

Already included in `shp-tests.hpp`

lslusarczyk: It is not included (only view is).
And I need `expect_fast_transform_path` to be defined before including `dr/shp/algorithms/transform.hpp` to allow extra checks.

I'd like to split shp-tests.hpp and create some header which will contain only test common stuff without including all shp code. Then use this header here, include only needed stuff.

test/gtest/shp/transform.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+template <typename AllocT> class TransformTest : public testing::Test {
+public:
+  using DistVec = shp::distributed_vector<typename AllocT::value_type, AllocT>;
+  using LocalVec = std::vector<typename AllocT::value_type>;
+  constexpr static const auto add_10_func = std::bind(
+      std::plus<typename AllocT::value_type>(), std::placeholders::_1, 10);
```
rscohn2, line 17: You must have tried...

Why not something simpler like functor, template function/lambda?

include/dr/shp/distributed_vector.hpp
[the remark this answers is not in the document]
rscohn2: @BenBrock: Unaligned requires cross-GPU references, and that gets into coherency issues. What is the SHP memory model?

test/gtest/shp/transform.cpp
rscohn2: shp.hpp should include everything. User programs do not have individual includes.

include/dr/details/alignment.hpp
```diff
@@ -0,0 +1,33 @@
...
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "dr/concepts/concepts.hpp"
+
+namespace lib {
+namespace __details {
+
+// It won't have segments if it is a zip of non-aligned data
+bool aligned(distributed_iterator auto iter) {
```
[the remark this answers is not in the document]
lslusarczyk: > Peer2peer is an optional feature

Does calling `sycl::queue(...).memcpy(...)` between devices mean that we use optional peer2peer feature? (If so, then I understand I shall transfer by memcpy a memory from one GPU to host and then from host to another GPU)

Or does peer2peer mean reading/writing on some device an address allocated on another device? (which I was able to do on PVC - it just worked slower).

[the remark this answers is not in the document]
rscohn2: I expect queue::memcpy to copy through host on systems that do not have p2p, and do a direct copy otherwise. Load/store instructions do not work on systems without p2p. I don't know the PVC memory model, but expect that 2 GPU's accessing the same cache line (with at least one write) will have undefined behavior. We can optimize for PVC, but DR must work on all GPU's.

We need to have something simple that always works (all work on single GPU or the host), something that handles the easy case (aligned operands). I would not spend any time on the in-betweens for now. We don't want to put that code in every algorithm and we need use cases.

Ben needs to walk us through how he expects it should work. If Transform relies on for_each for parallelization, it will be more manageable.

include/dr/shp/distributed_vector.hpp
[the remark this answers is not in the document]
BenBrock: With `shp` I'm assuming that peer-to-peer access is enabled and that GPUs can access each others' memory in a cache-coherent manner.  I believe this should be the case if you use USM shared on a multi-PVC system.  @jngkim maybe you can comment on the cross-GPU PVC memory model?

include/dr/shp/algorithms/transform.hpp
BenBrock: On second thought, I actually think the first implementation ought to perform well, as the memory controller should be able to properly handle the cross-GPU copies.  (At least I would expect this to perform well with P2P on Nvidia GPUs.)

include/dr/details/alignment.hpp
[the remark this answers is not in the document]
BenBrock: We should rely on peer-to-peer.  It's a pretty basic and ubiquitous feature.  Peer-to-peer works on virtually all Nvidia multi-GPU systems---it only requires that GPUs are on the same PCIe root bus or are connected with NVLink.  Peer-to-peer doesn't currently work with SYCL on any Nvidia systems only because the open-source DPC++ implementation hasn't added support yet.  (But it should be if SYCL on Nvidia is ever expected to be used in production.)

I don't think we should handicap our system based on a temporary driver limitation.

It is worth checking whether peer-to-peer currently works with Intel multi-GPU systems that only connect GPUs over PCIe.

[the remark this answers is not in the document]
rscohn2: AWS P4 instances cannot read/write peer memory. I see that geforce does not either. They are affordable multi-gpu development systems. It might be product segmentation. We could defer during POC, but unless availability changes I think we need a working fallback for the long term.

include/dr/shp/algorithms/transform.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+    sycl::queue q(shp::context(), policy.get_devices()[in_seg.rank()]);
+    const std::size_t seg_size = rng::size(in_seg);
+    assert(seg_size == rng::size(out_seg));
+    auto local_in_seg = __detail::get_local_segment(in_seg);
+
+    if (in_seg.rank() == out_seg.rank()) {
+      auto local_out_seg = __detail::get_local_segment(out_seg);
+      events.emplace_back(
+          q.parallel_for(seg_size, [=](auto idx) { local_out_seg[idx] = fn(local_in_seg[idx]); }));
```
rscohn2, line 39: You are doing a lambda capture of something that has pointers? I don't think that will work.

rscohn2:
> do you have some hints how to debug failures like one below when running on PVC GPU?

I have seen that problem when I pass an invalid pointer as a kernel argument.

lslusarczyk:
> > do you have some hints how to debug failures like one below when running on PVC GPU?
> 
> I have seen that problem when I pass an invalid pointer as a kernel argument.

Thank you, I'll check it.

Currently I see only a very basic ways of debugging sycl code on PVC: commenting code and rerunning it to narrow problem, do code review. There is no logging possible on PVC, no backtraces of failures. I wonder if there is any other way to trace such exceptions to see what operation they came from.

rscohn2:
Did you try the gdb that is part of oneapi? I never had any luck with it
and always did what you are doing.

include/dr/shp/algorithms/transform.hpp
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+#pragma once
+
+#include <dr/shp/init.hpp>
+#include <dr/shp/util.hpp>
+
+namespace shp {
+
+template <class ExecutionPolicy>
+void transform(ExecutionPolicy &&policy, lib::distributed_range auto &&in,
```
rscohn2, line 12: transform returns a result
https://en.cppreference.com/w/cpp/algorithm/ranges/transform

lslusarczyk: applied

include/dr/shp/algorithms/transform.hpp
lslusarczyk: I wrote it this way and found problems with using `shp::for_each`. I was loosing reference when zipping/unzipping `in/out` and `out` was finally not modified. I gave up being not sure if when I finally will make it working it will became more readable and as fast a current version, which may be easier for compiler to optimize.

But I don't object if anyone wants to try to use `shp::for_each` with this code - all UTs are already here, pass in orlab, so refactoring should be safe.

lslusarczyk:
`device_ref::operator T()` creates sycl::queue without shp::context(), just with default constructor.
Can then `distributed_vector::operator[]` it work correctly?

As status for today I am facing strange behaviour in tests when running on GPU machines in the lab. Yesterday code was passing all the time, like today morning. I even made copy bug look like be fixed and disabled tests were passing. Then one machine was giving success while other failures, then all machines were failing my tests, some executing tests very slowly. CPU works well all the time. I am getting more and more confused what is wrong, our code or drivers or hardware...

rscohn2:
> I am getting more and more confused what is wrong, our code or drivers or hardware...

Likely all of the above :)

lslusarczyk:
> What failures were you were getting? Some of the ORTCE machines do break from time to time, so if they suddenly started breaking it could be related.

(1) some test fail with "C++ exception with description "Native API failed. Native API returns: -999 (Unknown PI error) -999 (Unknown PI error)" thrown in the test body."
(2) some with junk values being read: "Expected equality of these values: b[1000] Which is: 21474836485 expected 17"
(3) some with wrong value being read: "Expected equality of these values: b[1001] Which is: 5 expected 17"
(4) some machines can not execute "shp-tests" in slurm time limit (normally shp-tests takes a few seconds to run): "slurmstepd: error: *** STEP 122764.0 ON sdp5259 CANCELLED AT 2023-03-21T08:59:53 DUE TO TIME LIMIT ***"

I will check tomorrow if fixing operator() in device_ref will fix failures 1-3

If it is going to be a usable product I think we need 

1. a fixed PVC setup(s) we target our code for, with defined set of drivers (agama/l0/I don't know what's more there) to stabilize tests on it, have repeatable failures, analyse each issue and pass it to right team or fix our code. Be able to get driver fixes quickly (or any other software stack we rely on). And run CI on it.
2. know contract (memory model, ...) between our code and sycl/l0 valid on our hardware setup
3. somehow deal with hardware issues - workarounding them, or disabling some of our functionality or changing hardware if new version of hardware fixes something

If this is not possible, then will it be possible to make product of shp that runs on GPUs? Or shall we focus on mhp? Can mhp/MPI utilize GPU cards and their memory by running one process per GPU?

lslusarczyk:
steps to reproduce issue

1. checkout branch of this PR
2. compile shp-tests binary on ortc and run on some gpu machines
```
module load intel/oneapi/2023.0.0
MKL_DIR=/opt/intel/oneapi/mkl/2023.0.0/lib/cmake/mkl CXX=icpx ~/usr/bin/cmake -DENABLE_SYCL:BOOL=ON -DENABLE_SYCL_MPI:BOOL=ON -S . -B build-icpx
CXX=icpx ~/usr/bin/cmake -DENABLE_SYCL:BOOL=ON -DENABLE_SYCL_MPI:BOOL=ON -S . -B build-icpx
cd build-icpx/test/gtest/shp
srun -p QZ1J-SPR-PVC-2C -w sdp5259 ./shp-tests
srun -C GPU_QZ1B ./shp-tests
srun -C GPU_QZ2G ./shp-tests
srun -C GPU_QZ2G ./shp-tests
srun -C GPUCT_2 ./shp-tests
```

PR #123, opened 2023-03-17
mhp::distributed_matrix initial version

mateuszpn:
Preliminary implementation of mhp::distributed_matrix + very simple example (stencil-2d-matrix). No halo yet.


include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,174 @@
...
+      if ((*_titr).is_local()) {
+        for (size_t s = 0; s < (*_titr).shape()[0]; s++) {
+          row_view_.emplace_back(data_ + s * (*_titr).shape()[1], *_titr, (*_titr).shape()[1]);
+        }
+      }
+    };
+    return row_view_;
+  }
+
+  std::vector<dm_row_view<T>> rows() {
```
rscohn2, line 88: rows() should return something lightweight that can be copied. There can be a large number of rows. It needs to be a span or view. I think this matches shp so something to discuss with @BenBrock

include/dr/mhp/containers/index.hpp
rscohn2, line 1: Is this a copy of something in SHP, with only a change in namespace?

rscohn2: Direct use of mdspan is not a solution because it does not provide iterators. It may be possible to leverage mdspan indexing. I heard that latest mdspan supports indexing like this: [x,y,z]. mdspan has flexible layout row-major order, col major order, etc. It scales to > 2 dimensions. We need to understand the tradeoffs between custom index solution and using mdspan concepts. Something to work out with @BenBrock

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,174 @@
...
+    return row_view_;
+  }
+
+
+
+  distributed_dense_matrix(key_type shape, lib::halo_bounds hb = lib::halo_bounds(), Allocator allocator = Allocator())
+      : shape_(shape), // partition_(new mhp::block_cyclic()),
+      grid_shape_(default_comm().size(), 1),
+      tile_shape_(( shape_[0] + default_comm().size() - 1)/default_comm().size(), shape_[1]),
+      segment_size_( std::max({ tile_shape_[0] * tile_shape_[1], hb.prev * tile_shape_[1], hb.next * tile_shape_[1] } )),
```
rscohn2, line 106: hb.prev affects tile_shape. A tile must be at least as big as a halo in all dimensions. halo needs to be extended to 2 dimensions. Since you don't have any of that, you should delete the halo and add it back in when you are ready to fully implement halo.

include/dr/mhp/containers/distributed_dense_matrix.hpp
rscohn2, line 1: I don't see the implementation of remote memory read/write. Using distributed_vector underneath would give you that.

include/dr/mhp/containers/index.hpp
mateuszpn: Yes, the code comes from shp matrices, and I copied it to have the same way of handling indices. Maybe we can consider moving common elements to lib::

PR #131, opened 2023-03-22
Revert to using default queue

BenBrock:


BenBrock:
Unfortunately changing the default queue to use the global context seems to introduce random failures in the CI, so reverting to use the default queue.

PR #132, opened 2023-03-22
added more shp::iota tests

haichangsi:


include/dr/mhp/algorithms/algorithms.hpp
```diff
@@ -116,16 +116,16 @@ void for_each(DI first, DI last, auto op) {
 //
 
 /// Collective iota on iterator/sentinel for a distributed range
-template <lib::distributed_iterator DI>
-void iota(DI first, DI last, auto value) {
+template <lib::distributed_iterator DI, std::integral T>
```
rscohn2, line 119: integral is too strong. Won't iota work for float? It is implementable for any type, but some types will not permit parallel execution.

BenBrock: The ranges version of `iota` actually doesn't work for float, since `weakly_incrementable` requires that the difference type be an integer-like type.  (For float it's a float, which is not integer-like.)

We could implement a more generic version, but it would be a lot of work.  Will people really want to use `iota` with things that are not integers?

`std::integral` is more restrictive than `std::weakly_incrementable`---the main thing I need from `integral` is the ability to add, rather than just increment, which is necessary for parallelism.  (`x + 12` rather than just `++x`)

lslusarczyk: I made it consistent with shp. @rscohn2, please decide if I should revert `std::integral` in mhp. If revert, then we need to have tests with floats too in mhp.

haichangsi:
> @haichangsi Can you get this branch up-to-date with `main` and then send it to me to review?

@BenBrock Lukasz is making some changes atm so it's still not ready for a review.

include/dr/mhp/algorithms/algorithms.hpp
rscohn2: `std::integral` make sense. It should be possible to implement iota on top of `xhp::copy` or `xhp::for_each`. They only support random access iterators, which is necessary for parallelism, so iota could only support integral.

Long term, I think we need to support non random access iterators for `xhp::for_each`. It won't be parallel, but`std::for_each` would be doing element-wise remote memory operations, which is probably 1000x slower.

lslusarczyk:
new iota Large test fails in mhp with bugs like 
```
Expected equality of these values:
  v[33000]
    Which is: 44377
  33000
```
I analysed it and reproduced this issue locally and it is due to bug in old intel MPI (the one from 2023.0 oneAPI). Problem vanished if I use non-Intel MPI or 2023.1 intel-MPI.

@rscohn2 , could you please update oneAPI which our CI runs in or teach me how to do it? I think the version is installed by `uses: rscohn2/setup-oneapi@v0` [line in ci.yml](https://github.com/oneapi-src/distributed-ranges/blob/main/.github/workflows/ci.yml#L89). Where is located the code, which this line uses?

rscohn2:
The icpx build is probably because it is using gcc10 parallel stl with new TBB, which has deprecated APIs that it uses. You need to use gcc12 STL by using -DGCC_TOOLCHAIN with cmake. I forgot what path it needs and am checking now

include/dr/mhp/algorithms/algorithms.hpp
BenBrock: There will potentially be iterators that are not exactly random access but support parallelism, so I agree that we should think about supporting non-random access iterators in the future.  (writing code generically and using `rng::distance`, etc.)

For example, I could imagine implementing a hash table---while the hash table iterator would not exactly be random access, we could iterate through the hash table in parallel on the GPU/across multiple processes.  (e.g. by splitting up the range recursively or implementing a custom `for_each` for the iterator type.)

In the meantime, though, I think it's reasonable to restrict to random access, since all of the data structures we're currently considering have random access (-ish) iterators, including the sparse matrix.

lslusarczyk:
I confirm our code was failing with all: old IntelMPI, new IntelMPI and MPICH (I was wrong it passes with new IMPI and MPICH, because of too few executions which I tested).

Bug was in our code. We were not waiting for local completion of MPI_Put, but modifying (incrementing values in it) send buffer. This is why some cells were receiving too high values (at the time of real buffer read loop of aiota was ahead).

Specification of MPI says about it in [section 12.3](https://www.mpi-forum.org/docs/mpi-4.0/mpi40-report.pdf#section.12.3) and [section 3.7.3](https://www.mpi-forum.org/docs/mpi-4.0/mpi40-report.pdf#subsection.3.7.3).

After i changed MPI_Put to MPI_Rput/MPI_Wait problem vanished (I hope now for real).

I wrote an exposing & verifying simple pure-MPI code for this problem: [mpi_put_problem.cpp.txt](https://github.com/oneapi-src/distributed-ranges/files/11178593/mpi_put_problem.cpp.txt) which exposes it well

PR #134, opened 2023-03-23
Use `shp::devices()` inside algorithms, disable non-aligned

BenBrock:
`inclusive_scan` tests


BenBrock:
Modify algorithms to use `shp::devices()` instead of the `devices` inside `device_policy`, which is currently unused.

Also disable non-aligned `inclusive_scan` tests [oneDPL #854](https://github.com/oneapi-src/oneDPL/issues/854) is resolved.

PR #135, opened 2023-03-23
Implement ranges version of `shp::fill`

BenBrock:


BenBrock:
Implement ranges version of `shp::fill`.

Also, introduce `shp::__detail::combine_events()`, which can be used to combine multiple SYCL events into one.

Ensured most `.submit()`s are capturing by value.

rscohn2:
@lslusarczyk, @BenBrock: The way the fill implementation is structured seems like a good pattern to implement algorithms:

* sync/async distributed range, built on top of:
* sync/async remote range

Some questions:

* sync fill returns std's return value. async fill does not. It looks like the return values for most of the algorithms are supporting chaining. They may not be useful for async.
* Fill async API's are SYCL-specific. MHP needs to support async. Ben had some ideas on utilizing native std C++ support.
* Do we want the remote range primitives to be part of public API? Are they useful as building blocks for people writing their own algorithms? We would need tests/doc for them, too.

lslusarczyk:
> * Do we want the remote range primitives to be part of public API? Are they useful as building blocks for people writing their own algorithms? We would need tests/doc for them, too.

What does it mean on some example? Do I understand correctly that an example is already implemented in this PR `lib::remote_contiguous_range R> sycl::event fill_async(R &&r, T value)` that takes `remote_range` as parameter?

In my opinion - yes because a remote range has a rank associated with it and iterators don't and people can only write effective code when they know ranks and can pass them to our algorithms.
If using iterators is incorrect or inefficient (what I'm asking about in previous comment) then I think even iterator interface should be removed to enforce using efficient way.

PR #137, opened 2023-03-23
Fix use of device iterators with `oneDPL`

BenBrock:


examples/shp/dot_product_benchmark.cpp
rscohn2, line 1: Please enable build & short running test:
https://github.com/oneapi-src/distributed-ranges/blob/ca4be0a9edd4593480ee6cd9016adfee0d65795c/examples/shp/CMakeLists.txt#L19
You may want to enable inclusive_scan example if you believe it will pass reliably.

BenBrock:
We'll have to hold on adding this example until [#855](https://github.com/oneapi-src/oneDPL/issues/855) and [#854](https://github.com/oneapi-src/oneDPL/issues/854) are resolved.

include/dr/details/segments_tools.hpp
```diff
@@ -200,7 +207,8 @@ auto segments_(V &&v) {
 }
 
 template <rng::range V>
-  requires(lib::is_zip_view_v<std::remove_cvref_t<V>>)
+  requires(lib::is_zip_view_v<std::remove_cvref_t<V>> &&
+           distributed_zip_reference<rng::range_reference_t<V>>)
```
BenBrock, line 211: @rscohn2 Is this the right condition for `segments` for `rng::zip_view`?

(Segments need to be disabled if it's not a zip of distributed ranges.)

rscohn2: I deleted this code in the main branch today. The new code uses a member function, not an ADL.
https://github.com/oneapi-src/distributed-ranges/blob/db40f4c266ee993a0c8f6c606c3d8fd30463a4cc/include/dr/mhp/views/zip.hpp#L30

Issue #82, opened 2023-03-04
view:zip


rscohn2:

rscohn2:
mhp works with `rng::views::zip` shp works with `shp::views::zip`
shp zip does not use tuple for zip 2.
Can a custom zip be implemented like transform?

rscohn2:
Test implemented in #81

BenBrock:
It's pair that I need to switch to for a zip with two elements, right?  (And I'm currently using tuple?)

rscohn2:
Right

PR #139, opened 2023-03-24
zip sycl for_each

rscohn2:
Test sycl zip for_each. MHP is broken because we use DPL for_each, and try to convert the iterators to raw pointers, but a zip has no raw ponter. SHP is not tested because of the tuple/pair mismatch.

Relates to #82

### DIFF
--- a/test/gtest/include/common/zip.hpp
+++ b/test/gtest/include/common/zip.hpp
@@ -25,8 +25,8 @@ TYPED_TEST(Zip, Zip3) {
 }
 
 auto zip_inner(auto &&r1, auto &&r2) {
-  return rng::views::zip(rng::subrange(r1.begin() + 1, r1.end() - 1),
-                         rng::subrange(r2.begin() + 1, r2.end() - 1));
+  return zhp::zip(rng::subrange(r1.begin() + 1, r1.end() - 1),
+                  rng::subrange(r2.begin() + 1, r2.end() - 1));
 }
 
 TYPED_TEST(Zip, Subrange) {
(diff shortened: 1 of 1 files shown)


rscohn2:
@BenBrock: I have been converting iterators to pointers before passing to DPL:
https://github.com/oneapi-src/distributed-ranges/blob/76f46c2cbf3099735c3318033e0de49e24d6bca1/include/dr/mhp/algorithms/algorithms.hpp#L69
How do I mark my iterators so there is no sycl buffer copy?

BenBrock:
> @BenBrock: I have been converting iterators to pointers before passing to DPL:
> 
> https://github.com/oneapi-src/distributed-ranges/blob/76f46c2cbf3099735c3318033e0de49e24d6bca1/include/dr/mhp/algorithms/algorithms.hpp#L69
> 
> 
> How do I mark my iterators so there is no sycl buffer copy?

Sorry for the slow reply---there's an [example of how to do this here](https://github.com/oneapi-src/distributed-ranges/blob/dot-product-onedpl/examples/shp/dot_product_benchmark.cpp#L29).  Basically, you create `direct_iterator` wrappers and pass those in to oneDPL.

lslusarczyk:
> there's an [example of how to do this here](https://github.com/oneapi-src/distributed-ranges/blob/dot-product-onedpl/examples/shp/dot_product_benchmark.cpp#L29).

the link does not work to me

rscohn2:
> the link does not work to me

https://github.com/oneapi-src/distributed-ranges/blob/20e3010cf1625b5d2a88a372e8ba54ba4342069b/examples/shp/dot_product_benchmark.cpp#L29

rscohn2:
Superseded by #157

Issue #148, opened 2023-03-28
DistributedVector.ConstructorFillAOS failed


lslusarczyk:

rscohn2:
> I am rerunning

It passed on rerun.

lslusarczyk:
> It passed on rerun.

So the bug if the worse kind - a non-deterministic bug, which may require thousands or reruns to show up again.

> This is possibly related. https://github.com/oneapi-src/distributed-ranges/issues/125#issuecomment-1483197278

I think reason is different there. I [commented](https://github.com/oneapi-src/distributed-ranges/issues/125#issuecomment-1486906667) about it there.

lslusarczyk:
I've found an issue with fill, that causes filled values not to be visible for some time after _fill_ at peer GPUs and reported it as [CMPLRLLVM-46438](https://jira.devtools.intel.com/browse/CMPLRLLVM-46438)

BenBrock:
Cool! Great to know the bug is not in our code.  Looks like a synchronization bug in our SYCL implementation.

BenBrock:
Closing this, as this has been fixed by #193.

Issue #125, opened 2023-03-18
gemv failures: Native API returns: -34 (PI_ERROR_INVALID_CONTEXT)


rscohn2:
Fails in CI. It does not fail every time. It does not fail when running manually.

[ RUN      ] SparseMatrix.Gemv
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -34 (PI_ERROR_INVALID_CONTEXT) -34 (PI_ERROR_INVALID_CON
 
Disabled in CI with #124 

Looks similar to #121

rscohn2:
@BenBrock @lslusarczyk 
Another sighting in shp-3-only: https://github.com/oneapi-src/distributed-ranges/actions/runs/4511947761/jobs/7944746094
```
[----------] 4 tests from DistributedVector
[ RUN      ] DistributedVector.ConstructorBasic
[       OK ] DistributedVector.ConstructorBasic (28 ms)
[ RUN      ] DistributedVector.ConstructorFill
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -34 (PI_ERROR_INVALID_CONTEXT) -34 (PI_ERROR_INVALID_CONTEXT)
<end of output>
Test time =   9.03 sec
----------------------------------------------------------
Test Failed.
"shp-3-only" end time: Mar 24 14:07 UTC
"shp-3-only" time elapsed: 00:00:09
```

Here is the test: https://github.com/oneapi-src/distributed-ranges/blob/76f46c2cbf3099735c3318033e0de49e24d6bca1/test/gtest/include/common/distributed_vector.hpp#L54

lslusarczyk:
since #131 `sycl::queue` is sometimes given `sycl::contex` we create in `init.hpp` by `internal::global_context_ = new sycl::context(internal::devices_);` and sometimes just default context.

This suspicious to me. @BenBrock , AFAIR you said that default context should be equal to the one which we created. But this is not obvious to me. Could we stick to always use our context created by `init.hpp`, which should make things clear that context is indeed the same in all queues we use?

BenBrock:
@lslusarczyk I switched to always using our context, and that worked fine on ORTCE, but it broke a bunch of things in the CI, which is why I reverted back.  Happy to try switching back if you guys want to.

At this point I feel like there must be a bug in the SYCL CPU runtime that the CI somehow triggers.

[...]

Issue #147, opened 2023-03-28
MHP SYCL support


lslusarczyk:
Support 1 GPU/rank in MHP model, using SYCL. See https://github.com/oneapi-src/distributed-ranges/blob/old-mpi-model/include/dr/algorithms/sycl_algorithms.hpp for a way it can be done. Need to unify with SHP in the management of GPUs.

lslusarczyk:
@rscohn2 , do I understand correctly that this is a feature which is beyond our current OKRs and plan for 2023?

rscohn2:
See https://github.com/oneapi-src/distributed-ranges/blob/db40f4c266ee993a0c8f6c606c3d8fd30463a4cc/include/dr/mhp/algorithms/algorithms.hpp#L65 for an example in MHP. It uses execution policy to decide whether to execute on GPU or CPU, which is possible when using USM shared. The old code used data type, recognizing SYCL allocators. 

I do not expect much interest in something that cannot target GPU and intend to support MHP/GPU for 2023. Currently, the most popular way to scale to multiple GPUs is through multi-process.

lslusarczyk:
This is important info, that MHP will work on GPU and this is the main target. It makes sense. Then this is a very important task. But due to this I have observations:
- Compiling & stabilizing with gcc has little sense - it will work on CPU only.
- we don't test mhp/sycl version, we develop CPU version hoping sycl one will start to work slightly - I would change testing from CPU to GPU. Practice from shp shows that obvious things are failing (like equality checks failing but printing out having right values a moment later).
- why to have such "if constexpr"? Let's remove whole CPU code branches, stick to SYCL version only and stabilize it. It can also work on CPU, just like shp does, by using CPU sycl device. Why to have 2 ways of doing the same (CPU branch of code, SYCL branch running on CPU)? The only reason is to support g++, but if it is going to be of little interest, then I would drop it, stick to icpx/sycl only.
- we've just disabled usm_shared allocator from shp because of its stability issues. But I hope it will be easier to have it stable when one device is used in the process. But it is worth to start testing to have time to make it stable.

rscohn2:
My main concern about being SYCL-only is productivity because of debuggability.  I am not ready to disable non-sycl CPU target, but I can try to switch to SYCL-first. Direct_iterators fixed my issues with sycl/zip/for_each, so it should be easy to bring up the rest of the algorithms on top on that. SYCL-only means I can match SHP async APIs.

[...]

PR #155, opened 2023-03-29
consistent way of using queues in shp

lslusarczyk:
As I discussed with Ben
- code assumes that devices in device_policy are the same as global devices, therefore there is no need, and is error-prone to have devices in device_policy
- in order to narrow where is root cause of https://github.com/oneapi-src/distributed-ranges/issues/148 and https://github.com/oneapi-src/distributed-ranges/issues/125 bugs we need to have consistent way of using context and devices in shp

This PR is intended to make use of sycl::queue consistent in shp, stop making experimens on how we construct queues in various places. When we use one way we think should be correct, then we will be able to go to oneDPL/L0 to find out root cause.

### DIFF
--- a/include/dr/shp/algorithms/fill.hpp
+++ b/include/dr/shp/algorithms/fill.hpp
@@ -18,8 +18,9 @@ template <typename T, lib::remote_contiguous_range R>
 sycl::event fill_async(R &&r, const T &value) {
   // fill can not be directed to node not owning memory being filled
   auto q = __detail::queue_for_rank(lib::ranges::rank(r));
-  return q.fill(std::to_address(rng::begin(lib::ranges::local(r))), value,
-                rng::distance(r));
+  T *arr = std::to_address(rng::begin(lib::ranges::local(r)));
+  return q.parallel_for(sycl::range<1>(rng::distance(r)),
+                        [arr, value](sycl::id<1> idx) { arr[idx] = value; });
 }
 
 template <typename T, lib::remote_contiguous_range R>
--- a/test/gtest/shp/containers.cpp
+++ b/test/gtest/shp/containers.cpp
@@ -26,14 +26,12 @@ TYPED_TEST(DistributedVectorTest,
   EXPECT_EQ(second[0], lib::ranges::segments(second)[0][0]);
 }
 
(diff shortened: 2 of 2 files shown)


rscohn2:
mhp is similar in that there is only 1 possible GPU to use for each rank. But we support CPU and GPU and I have been using policy to decide where to execute. Another possibility is to inspect the allocator and execute on GPU when using USM device/shared memory. In that case, I could get rid of the policy argument. Any thoughts on that?

lslusarczyk:
> mhp is similar in that there is only 1 possible GPU to use for each rank. But we support CPU and GPU and I have been using policy to decide where to execute. Another possibility is to inspect the allocator and execute on GPU when using USM device/shared memory. In that case, I could get rid of the policy argument. Any thoughts on that?

I proposed [here](https://github.com/oneapi-src/distributed-ranges/issues/147#issuecomment-1488811248) to have just sycl branches and execute sycl code on CPU or GPU by selecting right device in sycl. There will be no policy argument and no inspecting allocator needed too :)

include/dr/shp/algorithms/inclusive_scan.hpp
```diff
@@ -38,18 +38,16 @@ void inclusive_scan_impl_(ExecutionPolicy &&policy, R &&r, O &&o,
 
     std::vector<sycl::event> events;
 
-    auto root = shp::devices()[0];
-    shp::device_allocator<T> allocator(shp::context(), root);
+    sycl::queue root = __detail::default_queue();
```
BenBrock, line 41: For operations like allocation we should actually create the queue on a particular device, so this needs to be reverted.

lslusarczyk: > For operations like allocation we should actually create the queue on a particular device, so this needs to be reverted.

My change did not changed device or context in this place. It was `shp::devices()[0]` and it is `default_queue` which is the same device. Context also is the same. If so - revert fixes nothing. Correct me if I'm wrong.

lslusarczyk:
change which was tried to be done here was already implemented in https://github.com/oneapi-src/distributed-ranges/pull/193

PR #161, opened 2023-03-31
cleanup header/namespaces

rscohn2:
`include ""` -> `include <>`
`_detail`, `internal` -> `__detail`
`#include <dr/details/...>` -> `#include <dr/detail/...>`

eliminate current directly relative include

add some enforcement with dr-style.py, but only gating commit for `dr/include/mhp`

@BenBrock @lslusarczyk


rscohn2:
@BenBrock: Can you look at the icpx build failure? DPL has some constraint that I do not understand.

PR #167, opened 2023-04-03
Add shp enumerate gtest

suyashbakshi:


rscohn2:
It's a good test. I will reuse your checking code when I test zip and want to modify an element.

rscohn2:
Superseded by #183

PR #174, opened 2023-04-04
using parallel_for instead of fill because of bug in fill

lslusarczyk:
workaround fill by not using queue fill, but parallel_for without kernel name
as described in CMPLRLLVM-46438
added unittests failing before and passing after the change

### DIFF
--- a/include/dr/mhp/containers/distributed_vector.hpp
+++ b/include/dr/mhp/containers/distributed_vector.hpp
@@ -122,7 +122,8 @@ public:
   auto rank() const { return segment_index_; }
   auto local() const { return dv_->data_ + index_ + dv_->halo_bounds_.prev; }
   auto segments() const {
-    return lib::internal::drop_segments(dv_->segments(), index_);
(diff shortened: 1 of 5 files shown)


lslusarczyk:
> Unless you want to resolve the mhp issue, I can look at it later today.

Thank you for being ready to help. I think I found root cause and fixed it in the second commit of this PR.

rscohn2:
Still broken. Using main after you merged. Running `shp-tests` on anpfclxlin02.an.intel.com. It passed when I reran.
```
[----------] 4 tests from Fill/0, where TypeParam = shp::distributed_vector<int, shp::device_allocator<int, 0ul> >
[ RUN      ] Fill/0.Range
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -34 (PI_ERROR_INVALID_CONTEXT) -34 (PI_ERROR_INVALID_CONTEXT)
Aborted (core dumped)
(venv) rscohn1@anpfclxlin02:shp$
```

PR #176, opened 2023-04-05
iterators use <=>,==

rscohn2:
Convert some iterators to use <=>, == to define all relational operations

### DIFF
--- a/include/dr/details/iterator_adaptor.hpp
+++ b/include/dr/details/iterator_adaptor.hpp
@@ -63,30 +63,10 @@ public:
     return accessor_ == other.accessor_;
   }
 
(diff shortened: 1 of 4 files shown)


rscohn2:
Setting this aside for now. I have mhp working, but shp fails to compile.

BenBrock:
I need to look into this.  The spaceship operator greatly simplifies some things.

rscohn2:
I gave you write access to my fork in case you want to push updates to this branch.

I did make 2 attempts to enable SHP, but I ran into type issues that I did not understand

rscohn2:
Closing. @BenBrock noted that we may want to support iterators that do not implement `<=>`

PR #185, opened 2023-04-10
Update method of creating queues

BenBrock:


BenBrock:
Use a uniform method of creating queues.

PR #187, opened 2023-04-10
Some updates to make getting the local segment/iterator easier

BenBrock:


include/dr/details/ranges.hpp
```diff
@@ -164,6 +165,27 @@ struct local_fn_ {
 
 inline constexpr auto local = local_fn_{};
 
+namespace __detail {
+
+template <typename T>
+concept has_local = requires(T &t) {
+  { lib::ranges::local(t) } -> std::convertible_to<std::any>;
```
rscohn2, line 172: I am curious, why is the `-> std::convertible_to<std::any>` needed? What would be the issue if you did not have it?

PR #190, opened 2023-04-11
mhp/iota

lslusarczyk:


rscohn2 (approved):
Wondering if @BenBrock knows if there is some way with ADL or other to avoid having to include this in both mhp/shp. The way it is now if OK for me if this is the simplest way to do it.

It would be good if MHP algorithms like copy/transform/etc implemented in a similar way.

include/dr/details/algorithms.hpp
```diff
@@ -2,17 +2,14 @@
...
-namespace shp {
+// ALGORITHMS CODE SHARED BETWEEN shp AND mhp
+// DO NOT INCLUDE THIS FILE DIRECTLY
+// INCLUDED BY algorithms.hpp FILES IN mhp AND shp
 
 template <lib::distributed_range R, std::integral T> void iota(R &&r, T value) {
   auto iota_view = rng::views::iota(value, T(value + rng::distance(r)));
 
-  shp::for_each(shp::par_unseq, shp::views::zip(iota_view, r), [](auto &&elem) {
+  for_each(par_unseq, views::zip(iota_view, r), [](auto &&elem) {
```
rscohn2, line 12: I think I will be changing mhp execution policies, and changing to shp way of just having par_unseq.
I am still fixing some bugs in mhp::zip. Let me know if you see problems

lslusarczyk:
> Wondering if @BenBrock knows if there is some way with ADL or other to avoid having to include this in both mhp/shp. The way it is now if OK for me if this is the simplest way to do it.

I had the same question in mind and I'd like to verify it with you before merge if you have some better/simpler ideas.

> It would be good if MHP algorithms like copy/transform/etc implemented in a similar way.

Indeed. Do you know about any arguments/reasons against sharing zip view too? We have two complex implementations but it looks to me to be possible to make it be one. But maybe I don't know about something.

> I am still fixing some bugs in mhp::zip. Let me know if you see problems

iota tests pass and IMO iota tests are quite complete and good now, so current zip_view works for iota purpose

rscohn2:
> Do you know about any arguments/reasons against sharing zip view too?

It would require some refactoring to make it possible to share. I think the mhp zip is simpler and more self-contained, but there might be issues with shp.

* shp zip always uses tuple. mhp relies on underlying rng::zip, so it will use pair for 2, and tuple for everything else
* mhp segments() is an empty range if the components are not aligned. shp will build segments that include data from multiple ranks
* local() might need to be different, too.
* mhp zip does not work for some compositions of views and algorithms. I am looking at that.

I disabled the common zip tests for shp because of the pair/tuple mismatch.

lslusarczyk:
Here is comment from @BenBrock about this PR I'd like to share also with @rscohn2 

> Could you provide a bit more detail about what we're trying to do there?
> Personally I think it is not unreasonable to have the code duplicated in both shp and mhp
> It's a 4-5 line function, after all
> How many algorithms are we likely to have that share code between mhp/shp?
> We might also find that we want to write an optimized version of one of these shared algorithms in mhp/shp, in which case untangling the shared code would be a bit tedious

I was trying to write faster iota for mhp.
The more algorithms we cn share between mhp/shp the better. It means that abstractions of distibuted structures we made are good and mature. IMO zip should also be unified.

It would be perfect if most of algorithms could be shared like iota with just some building blocks (like for_each) being different. This is why I'd like to consult with you what is the best way of share code even if this is 4 lines now, because I hope there will be more code to be shared and it is good to choose best way of sharing at the begining.

Code duplication is generally a code smell due to less readability, higher cost of maintenance, risk of leaving bugs in one branch. It should be avoided unless there are good reasons to have it. Potential optimization which we could introduce in one branch in the future is IMO not a good reason to write non-shared code from the beginning or to abandon code unification. In case  of optimization I'm OK with paying a cost of untangling the shared code in the future, but only if the optimization is worth of introducing duplication.

rscohn2:
> It would be good if MHP algorithms like copy/transform/etc implemented in a similar way.

When I said this, I meant relying on `for_each` to do the parallelization instead of explicitly parallelizing all the algorithms.

The code is short and the way we share is a bit unnatural--including the same code in different namespaces. I change my vote to duplicate.

lslusarczyk:
> When I said this, I meant relying on for_each to do the parallelization instead of explicitly parallelizing all the algorithms.

I meant sharing code between mhp and shp is a good idea and I'd like to share what can be shared. I know not all can be.

> The code is short and the way we share is a bit unnatural--including the same code in different namespaces

Anyway, this was my concern too. Hence I asked for your opinions. Thank you for review.

I reverted sharing code by "include". I feel that we can invent some better way than includes, but let's postpone researching how to share algorithms until something bigger will be possible to be shared.

include/dr/shp/shp.hpp
```diff
@@ -9,8 +9,6 @@
 #include <dr/detail/ranges_shim.hpp>
 #include <dr/detail/segments_tools.hpp>
```
rscohn2, line 10: This is a general comment and not specific to this PR. For mhp.hpp, my intention is to remove everything that is not a direct dependence of a user program, so I would get rid of `ranges_shim.hpp`, `iterator_adaptor.hpp`, ...

It should mostly be everything in algorithms/containers/views directories + implementation of the CPO's like dr::ranges::segments and local.

lslusarczyk: Great! This is how mature libraries look like, e.g. [boost/filesystem.hpp](https://github.com/boostorg/filesystem/blob/fcc1bf3eca15323b90f2df042464668b0d131698/include/boost/filesystem.hpp) does not include any details. Details are included by particular headers that need them, e.g. [here in in boost/filesystem/directory.hpp](https://github.com/boostorg/filesystem/blob/fcc1bf3eca15323b90f2df042464668b0d131698/include/boost/filesystem/directory.hpp#L30).

@BenBrock, can we/I do the same in [shp.hpp](https://github.com/oneapi-src/distributed-ranges/blob/main/include/dr/shp/shp.hpp) that is 

- remove dr/detail/ from there 
- also move dr/shp/shp.hpp one level up to dr/shp.hpp in order to make it consistent with mhp, and includes style used e.g. by mentioned boost libs

?

rscohn2: > * also move dr/shp/shp.hpp one level up to dr/shp.hpp in order to make it consistent with mhp, and includes style used e.g. by mentioned boost libs

And oneapi style agrees: https://github.com/oneapi-src/distributed-ranges/blob/main/doc/developer/design/oneapi-cpp-style-guide.md#additional-recommendations-on-includes

PR #191, opened 2023-04-11
mhp::distributed_dense_matrix + stencil

mateuszpn:
mhp::distributed_dense_matrix class (rows distributed over MPI nodes) plus:
- halo over rows
- row manipulation class, and iterator over rows
- dedicated dr::mhp::subrange defined as subrectangle, with iterator
- dr::mhp::transform for the subrange
- a stencil, iterating over elements of the dr::mhp::subrange, added in stencil-2d-matrix example.


include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+namespace dr::mhp {
+
+using key_type = index<>;
+
+class distributed_matrix_partition {};
+class by_row final : public distributed_matrix_partition {};
+class by_column final : public distributed_matrix_partition {};
+class tiled final : public distributed_matrix_partition {};
+
+template <typename DM> class dm_segments : public std::span<d_segment<DM>> {
```
lslusarczyk, line 21: remove this class, it is not needed as you may use rng::all view like distributed_vector does, and it is also dangerous as it not derives from `rng::view_base`

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+  T *data_ = nullptr; // local data ptr
+
+  dr::span_halo<T> *halo_;
+  dr::halo_bounds halo_bounds_;
+
+  std::vector<d_segment<distributed_dense_matrix>> segments_;
+  dm_segments<distributed_dense_matrix>
+      dm_segments_; // lightweight view on segments_
+
+  dm_rows<distributed_dense_matrix<T>>
```
lslusarczyk, line 268: one time you use `distributed_dense_matrix` and another time `distributed_dense_matrix<T>`
Stick to one notation. The one which is better if one of them is better than other.

include/dr/mhp/containers/rows.hpp
```diff
@@ -0,0 +1,238 @@
...
+    // return lib::internal::drop_segments(dm_->segments(), index_);
+    return dm_->segments();
+  }
+  auto &halo() const { return dm_->halo(); }
+
+  bool is_local() { return dm_->is_local_row(index_); }
+
+private:
+  DM *dm_ = nullptr;
+  std::size_t index_ = 0;
```
lslusarczyk, line 210: `index_` may mean everything, `row_number` is more descriptive IMO

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+    hb.next *= shape_[1];
+
+    halo_ = new dr::span_halo<T>(default_comm(), data_, data_size_, hb);
+
+    // prepare segments
+    // one dsegment per node, 1-d arrangement of segments
+
+    segments_.reserve(grid_size_);
+
+    std::size_t idx = 0;
```
lslusarczyk, line 176: `idx` equals `i` all the time, one of them is useless

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+    for (std::size_t i = 0; i < grid_size_; i++) {
+      std::size_t _seg_rows =
+          segment_shape_[0] -
+          ((idx + 1) / default_comm().size()) *
```
lslusarczyk, line 180: ` ((idx + 1) / default_comm().size())` is always 0 except last segment so last segment gets rows - excessive rows

first - this can be written in more readable way than dividing, e.g. `(i+1 == grid_size_ ? ... : 0 )` or simple `if`
second - not only last segment may be empty but several last segments can be (6 rows 5 processes, 1 row 3 processes)

lslusarczyk (changes requested):

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+  distributed_dense_matrix(
+      key_type shape, dr::halo_bounds hb = dr::halo_bounds(),
+      distributed_matrix_partition *partition = new by_row(),
+      Allocator allocator = Allocator())
+      : shape_(shape), segment_shape_((shape_[0] + default_comm().size() - 1) /
+                                          default_comm().size(),
+                                      shape_[1]),
+        segment_size_(std::max({segment_shape_[0] * shape_[1],
+                                hb.prev * shape_[1], hb.next * shape_[1]})),
```
lslusarczyk, line 68: as we discussed code does not support halo being greater than segment

therefore segment_size is always `segment_shape_[0] * segment_shape_[1]`
add assertion that `segment_shape_[0] <= max(hb.prev, hb.next)` with comment that this is not yet implemented

include/dr/mhp/containers/rows.hpp
```diff
@@ -0,0 +1,238 @@
...
+  using dsegment = d_segment<dmatrix>;
+
+public:
+  using iterator = typename std::span<T>::iterator;
+
+  dm_row(){};
+  dm_row(signed long idx, T *ptr, std::size_t size, dsegment *segment, 
+         Allocator allocator = Allocator())
+      : std::span<T>({ptr, size}), index_(idx), data_(ptr), segment_(segment),
+        size_(size), allocator_(allocator){};
```
lslusarczyk, line 31: `ptr` and `size` are already in `std::span`
no need to double it in dm_row

include/dr/mhp/containers/rows.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+  // copying ctor
+  dm_row(const dm_row &other)
+      : std::span<T>({(other.index_ == INT_MIN)
```
lslusarczyk, line 35: span is a view, not container, not intended to own memory
the same I expected dm_row to be

mem is allocated here twice - different in span and different in data, deallocated is only one of them in destructor

why dm_row need to own

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+  size_type size() noexcept { return shape()[0] * shape()[1]; }
+  auto segments() { return dm_segments_; }
+  size_type segment_size() { return segment_size_; }
+  key_type segment_shape() { return segment_shape_; }
+
+  auto &halo() { return *halo_; }
+  dr::halo_bounds &halo_bounds() { return halo_bounds_; }
+
+  bool is_local_row(int index) {
+    if (index >= local_rows_indices_.first &&
```
lslusarczyk, line 100: instead of
```
if (condition) return true; else return false
```
write `return condition`
Here and below.

include/dr/mhp/containers/segments.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+#pragma once
+
+#include "index.hpp"
+
+namespace dr::mhp {
+
+using key_type = index<>;
+
+template <typename DV> class d_segment_iterator;
```
lslusarczyk, line 14: looks the same as dv_segment_iterator, dv_segment_reference
what is different? I don't see any specific 2d code here
Can it be shared? if so please just use dv_segment_iterator and dv_segment_reference

include/dr/mhp/containers/rows.hpp
mateuszpn: Done

examples/mhp/CMakeLists.txt
```diff
@@ -23,6 +23,20 @@ target_link_libraries(
   DR::mpi)
 add_mpi_test(stencil-1d-array stencil-1d-array 2)
 
+# add_executable(stencil-2d stencil-2d.cpp)
```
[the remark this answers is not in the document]
mateuszpn: Done

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: To be verified, whether it is or it is not implemented in halo_impl class

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: 1st - changed
2nd - the case (not likely) protected by assert in init_()

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+#include "index.hpp"
+#include "segments.hpp"
+#include "rows.hpp"
+#include "subrange.hpp"
+
+namespace dr::mhp {
+
+using key_type = index<>;
+
+class distributed_matrix_partition {};
```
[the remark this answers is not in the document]
mateuszpn: OK, but we will come back to it sooner or later

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: Shorter is better, fixed

include/dr/mhp/containers/rows.hpp
mateuszpn: row_idx is a compromise, I suppose

include/dr/mhp/containers/segments.hpp
mateuszpn: TODO

include/dr/mhp/containers/index.hpp
```diff
@@ -0,0 +1,110 @@
+// SPDX-FileCopyrightText: Intel Corporation
```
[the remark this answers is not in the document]
mateuszpn: TODO

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,323 @@
...
+          segment_shape_[0] -
+          ((idx + 1) / default_comm().size()) *
+              (default_comm().size() * segment_shape_[0] - shape_[0]);
+      segments_.emplace_back(this, idx, _seg_rows * shape_[1]);
+      idx++;
+    }
+
+    // regular rows
+    dm_rows_.reserve(segment_shape_[0]);
+
```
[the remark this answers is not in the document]
mateuszpn: I can see no way of significant simplification

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+template <typename T> class MhpTests3 : public testing::Test {};
+
+TYPED_TEST_SUITE(MhpTests3, AllTypes);
+
+TYPED_TEST(MhpTests3, suite_works_for_3_processes_only) {
+  EXPECT_EQ(dr::mhp::default_comm().size(), 3); // dr-style ignore
+}
+
+TEST(MhpTests3, DM_Index) {
+  assert(dr::mhp::default_comm().size() == 3);
```
lslusarczyk, line 22: I don't think it is needed in every tests, since we have a separate test that checks it.
But I don't insist.

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+TEST(MhpTests3, DM_For) {
+  assert(dr::mhp::default_comm().size() == 3);
+  const int rows = 11, cols = 11;
+  DM a(rows, cols, -1);
+
+  for (auto i : a) {
+    i = 5;
+  }
+
+  if (dr::mhp::default_comm().rank() == 0) {
```
lslusarczyk, line 58: same comment as to previous test case

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+  } else /*(dr::mhp::default_comm().rank() == 2) */ {
+    EXPECT_EQ(a.begin()[0], 5);
+    EXPECT_EQ(a.begin()[15], 5);
+    EXPECT_EQ((a.begin()[{4, 4}]), 5);
+    EXPECT_EQ(a[99], 5);
+    EXPECT_EQ((a[{10, 10}]), 5);
+  }
+}
+
+TEST(MhpTests3, DM_Fill) {
```
lslusarczyk, line 79: same comment

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+  } else /*(dr::mhp::default_comm().rank() == 2) */ {
+    EXPECT_EQ(a.begin()[0], 0);
+    EXPECT_EQ(a.begin()[15], 15);
+    EXPECT_EQ((a.begin()[{4, 4}]), 48);
+    EXPECT_EQ(a[99], 99);
+    EXPECT_EQ((a[{10, 10}]), 120);
+  }
+}
+
+TEST(MhpTests3, DM_Iota2) {
```
lslusarczyk, line 162: same comment + request to use meaningful test names (instead of Iota2 better is to name e.g. Iota_on_subrange)

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+  } else /*(dr::mhp::default_comm().rank() == 2) */ {
+    EXPECT_EQ(a.begin()[0], -13);
+    EXPECT_EQ(a.begin()[15], 15);
+    EXPECT_EQ((a.begin()[{4, 4}]), 48);
+    EXPECT_EQ(a[99], 99);
+    EXPECT_EQ((a[{10, 10}]), -13);
+  }
+}
+
+TEST(MhpTests3, DM_Iota_HB) {
```
lslusarczyk, line 189: same comment (remove `if`, move to non-3-only test)

test/gtest/mhp/distributed_dense_matrix3.cpp
mateuszpn: and same answer...

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+TEST(MhpTests3, DM_Index) {
+  assert(dr::mhp::default_comm().size() == 3);
+
+  const int rows = 11, cols = 11;
+  DM a(rows, cols, -1);
+  fence();
+
+  if (dr::mhp::default_comm().rank() == 0) {
```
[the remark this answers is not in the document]
lslusarczyk: I understand now.
Change remote-access checks to be outside if
and inside if only do local-access check

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,492 @@
...
+  dr::rma_window win_;
+  Allocator allocator_;
+}; // class distributed_dense_matrix
+
+//
+// for_each, but iterator is passed to lambda
+//
+template <typename T>
+auto for_each(dm_rows<distributed_dense_matrix<T>> &rows, auto op) {
```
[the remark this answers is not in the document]
mateuszpn: This is not a quick and easy fix, it requires a deeper rewrite of dm_rows::segments and more. I suggest postponing this and adding it as a separate task for the (not too distant) future. May apply also to transform() below.

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,492 @@
...
+};
+
+template <typename T>
+concept has_segment = requires(T t) { (*(t.begin())).segment(); };
+
+template <typename T1, typename T2>
+  requires(has_segment<T1> && has_segment<T2>)
+auto for_each(zip_view<T1, T2> &v, auto op) {
+  fmt::print("{}: inside for_each \n", default_comm().rank());
+  for (auto itr = rng::begin(v); itr != rng::end(v); itr++) {
```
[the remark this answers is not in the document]
mateuszpn: This version of for_each is currently not used, still it should be fixed and restored to use. It is however not necessary in the current PR

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,492 @@
...
+  }
+};
+
+//
+// transform, but iterator pointing at element is passed to lambda instead of
+// element
+//
+template <typename DM>
+void transform(dr::mhp::subrange<DM> &in, subrange_iterator<DM> out, auto op) {
+  for (subrange_iterator<DM> i = rng::begin(in); i != rng::end(in); i++) {
```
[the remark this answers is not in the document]
mateuszpn: Substantial work, as above. Enabling SYCL should be included in a future task.

include/dr/mhp/containers/rows.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+  DM *dm() { return dm_; }
+
+private:
+  DM *dm_ = nullptr;
+  std::size_t row_idx_ = 0;
+}; // dm_rows_iterator
+
+template <typename DM>
+class dm_rows : public std::vector<dm_row<typename DM::value_type>> {
```
rscohn2, line 155: Why do you inherit from std::vector? You don't initialize it.
I think you want this to be range adaptor: a view of an underlying range.

include/dr/shp/containers/matrix_entry.hpp
```diff
@@ -17,14 +17,14 @@ template <typename T, typename I = std::size_t> class matrix_entry {
   using index_type = I;
   using map_type = T;
 
-  matrix_entry(dr::shp::index<I> index, const map_type &value)
+  matrix_entry(dr::index<I> index, const map_type &value)
```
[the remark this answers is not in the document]
rscohn2: I think you should do this as a separate commit. Big PR's that changes many files are hard to manage and hard to review.

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,496 @@
...
+    ++out;
+  }
+}
+
+} // namespace dr::mhp
+
+// Needed to satisfy rng::viewable_range
+
+template <typename T>
+inline constexpr bool rng::enable_borrowed_range<
```
rscohn2, line 486: You won't need this is dm_rows inherits from view_interface

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,496 @@
...
+    dr::mhp::dm_rows<dr::mhp::distributed_dense_matrix<T>>> = true;
+
+template <typename T>
+inline constexpr bool rng::enable_borrowed_range<
```
rscohn2, line 490: You won't need this if subrange inherits from view_interface

include/dr/mhp/containers/distributed_dense_matrix.hpp
```diff
@@ -0,0 +1,496 @@
...
+    dr::mhp::subrange<dr::mhp::distributed_dense_matrix<T>>> = true;
+
+template <typename T>
+inline constexpr bool rng::enable_borrowed_range<
+    std::vector<dr::mhp::d_segment<dr::mhp::distributed_dense_matrix<T>>>> =
```
rscohn2, line 495: a std::vector cannot be a borrowed range because the iterator is not valid if the range is destructed

rscohn2 (commented):
I don't understand how this works yet and only made comments on superficial aspects.

include/dr/mhp/containers/rows.hpp
mateuszpn: This is not a view, but a structure containing a description of rows. Each element of the dm_rows vector contains a pointer to the start of the row data, its index and size. The segments() function needs to be reworked to make dm_rows work properly as a distributed array (in for_each etc.), and to get rid of custom functions.

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: It was temporary when hunting compilation errors. Removed.

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: Subrange modified, so I removed the commented code

include/dr/mhp/containers/distributed_dense_matrix.hpp
mateuszpn: dm_rows is not a view, so such an inheritance would make little sense I suppose

examples/mhp/stencil-2d-matrix-rows.cpp
```diff
@@ -0,0 +1,181 @@
...
+      rng::iota(*r, 10);
+  }
+
+  // subranges of rows
+  auto in = rng::subrange(a.rows().begin() + 1, a.rows().end() - 1);
+  auto out = rng::subrange(b.rows().begin() + 1, b.rows().end() - 1);
+
+  for (std::size_t s = 0; s < steps; s++) {
+    dr::mhp::halo(in).exchange();
+    dr::mhp::transform(in, out.begin(), stencil_op);
```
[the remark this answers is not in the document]
mateuszpn: I agree, still I had problems with for_each and zip_view, so that an iterator is passed to the stencil, not ref to an element. dm_rows class requires some more work (see https://github.com/oneapi-src/distributed-ranges/pull/191#discussion_r1235111325)

include/dr/mhp/containers/rows.hpp
lslusarczyk: > The segments() function needs to be reworked

Is this described in some task and in some TODO in the code?

include/dr/mhp/containers/segments.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+  auto size() const { return size_; }
+
+  auto begin() const { return iterator(dv_, segment_index_, 0); }
+  auto end() const { return begin() + size(); }
+
+  auto operator[](difference_type n) const { return *(begin() + n); }
+
+private:
+  DV *dv_;
```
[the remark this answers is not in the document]
lslusarczyk: There is still no ` = nullptr` as it was before.

include/dr/mhp/halo.hpp
```diff
@@ -315,6 +315,16 @@ template <typename T, typename Memory = default_memory<T>> class span_group {
 };
 
 struct halo_bounds {
+  halo_bounds(std::size_t bound_ = 0, bool periodic_ = false) {
+    prev = next = bound_;
+    periodic = periodic_;
+  }
+  halo_bounds(std::size_t prev_, std::size_t next_, bool periodic_ = false) {
```
lslusarczyk, line 322: please remove default values from ctors

int is convertible to bool - when one writes halo_bounds(1,0) it is not clear for the compiler if we use first or second ctor

include/dr/mhp/halo.hpp
```diff
@@ -349,12 +359,12 @@ class span_halo : public span_halo_impl<T, Memory> {
...
                 "owned groups {}/{} first/last\n", comm.first(), comm.last());
-    if (hb.prev > 0 && (hb.periodic || !comm.first())) {
-      owned.emplace_back(span.subspan(hb.prev, hb.prev), comm.prev(),
+    if (hb.next > 0 && (hb.periodic || !comm.first())) {
+      owned.emplace_back(span.subspan(hb.next, hb.next), comm.prev(),
                          communicator::tag::halo_reverse);
     }
-    if (hb.next > 0 && (hb.periodic || !comm.last())) {
-      owned.emplace_back(span.subspan(rng::size(span) - 2 * hb.next, hb.next),
+    if (hb.prev > 0 && (hb.periodic || !comm.last())) {
```
lslusarczyk, line 366: Please don't do release this change in this PR. Please finish and merge https://github.com/oneapi-src/distributed-ranges/pull/324 PR first

include/dr/shp/containers/matrix_entry.hpp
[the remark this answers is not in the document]
lslusarczyk: Let's then do it in a separate PR before merging this one.

examples/include/data-utils.hpp
```diff
@@ -4,6 +4,14 @@
 
 #pragma once
 
+#ifdef DR_FORMAT
```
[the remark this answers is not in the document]
lslusarczyk: @mateuszpn , could you please apply it?

test/gtest/mhp/distributed_dense_matrix3.cpp
```diff
@@ -0,0 +1,214 @@
...
+  EXPECT_EQ((a.begin()[{4, 4}]), 48);
+  EXPECT_EQ(a.begin()[99], 99);
+  EXPECT_EQ((a.begin()[{10, 10}]), 120);
+
+  if (dr::mhp::default_comm().rank() == 0) {
+    EXPECT_EQ(a[0], 0);
+    EXPECT_EQ(a[15], 15);
+  } else if (dr::mhp::default_comm().rank() == 1) {
+    EXPECT_EQ((a[{4, 4}]), 48);
+  } else /*(dr::mhp::default_comm().rank() == 2) */ {
```
lslusarczyk, line 133: `assert(dr::mhp::default_comm().rank() == 2)`
is as descriptive as your comment, but self verifying in addition

here and in other cases

mateuszpn:
TODO (in next PRs):
- implement real segments() in dm_rows, then replace transform with for_each in stencil-2d-matrix-rows
- look at performace of operation (see results for wave-equation)

CMakeLists.txt
```diff
@@ -170,6 +170,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     add_subdirectory(examples/shp)
     add_subdirectory(test/gtest/shp)
     add_subdirectory(benchmarks/gbench/shp)
+    add_compile_options(-fsycl)
```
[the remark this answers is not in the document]
lslusarczyk: remove this flag, as it is redundant now

examples/include/data-utils.hpp
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <../../include/dr/detail/format_shim.hpp>
```
lslusarczyk, line 7: will `#include <dr/detail/format_shim.hpp>` not work?

lslusarczyk (approved):
2 cosmetics, LGTM and can be merged

CMakeLists.txt
```diff
@@ -168,7 +168,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     add_compile_options(-Wno-tautological-constant-compare)
     # DPL turns on -fopenmp-simd, triggering optimization warnings in parallel stl
     add_compile_options(-Wno-error=pass-failed)
-    add_compile_options(-Wno-error=unknown-pragmas)
+    add_compile_options(-Wno-unknown-pragmas)
```
rscohn2, line 171: Please leave it. -Wno-error was intentional

rscohn2 (approved):

examples/include/data-utils.hpp
mateuszpn: No, and I don't really understand why not.

rscohn2: Build with `make VERBOSE=1` and look at the compile line. Does it have a `-I` that points to the include directory? Maybe it is included in an example that does not otherwise use DR and it is not being added. It is OK to add DR as a link library even if it only uses this part.

PR #195, opened 2023-04-12
Use `std::pair` in `shp::zip_view`

BenBrock:


rscohn2:
Enumerate test was failing. There are some things I don't understand about the tests and I rewrote them to be more like the other tests. I did that in a separate PR #196, which has been merged. It adds mhp support for enumerate by copying shp. mhp works now. Then I rebased this branch to get the new enumerate test and tried again. It gets past the equality compare and now has trouble printing the zip view. gtest requires `<<` and `==` to be defined for the data types. I added support for stream operator and std::pair, but that is not enough. It looks like I have an ADL related issue. I pushed what I have so you can see it. mhp zip_view provides the same value type as the underlying zip implementation and that might be an easier path.

test/gtest/include/common-tests.hpp
```diff
@@ -322,4 +350,36 @@ inline std::ostream &operator<<(std::ostream &os,
...
+inline std::ostream &operator<<(std::ostream &os,
+                                const std::pair<T1, T2> &obj) {
+  os << fmt::format("{}", obj);
+  return os;
+}
+
 } // namespace DR_RANGES_NAMESPACE
+
+namespace std {
+
```
rscohn2, line 370: @BenBrock if the overloads are in std, do we still need the same thing in rng and shp?

lslusarczyk:
> Enumerate test was failing.

Check if [this bugfix](https://github.com/oneapi-src/distributed-ranges/pull/209) also fixed your issue.

BenBrock:
@rscohn2 I believe I've got everything fixed.  The last issue was with the iota tests, which pass in an unbounded iota range to `views::zip`, which `shp::zip_view` does not support.  I got around this by adding an `xhp::views::iota`, which in `shp` gets you a `rng::iota_view` bounded by `numeric_limits<W>::max()`, and in `mhp` is just `rng::views::iota`.

BenBrock:
If it looks good to you, please go ahead and merge this.

test/gtest/include/common-tests.hpp
lslusarczyk: Can we avoid putting things to "std" namespace?

test/gtest/shp/shp-tests.cpp
```diff
@@ -4,25 +4,37 @@
...
+#include "common/zip.hpp"
+
+#else
+
+using AllTypes = ::testing::Types<xhp::distributed_vector<int>,
+                                  xhp::distributed_vector<float>>;
 
 #include "common/all.hpp"
+#include "common/distributed_vector.hpp"
 // Issue with 2 element zips
```
lslusarczyk, line 21: What is this issue? Does shp::enumerate / shp::zip does not work for two elements? If so I'd like to report a bug for it to keep it in mind. If it is fixed let's remove the comment.

test/gtest/shp/shp-tests.cpp
```diff
@@ -4,25 +4,37 @@
 
 #include "shp-tests.hpp"
 
-using AllTypes = ::testing::Types<dr::shp::distributed_vector<int>,
-                                  dr::shp::distributed_vector<float>>;
+// Use this for shorter build time
```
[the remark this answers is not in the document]
rscohn2: I will disable the `MINIMAL_TEST` before merging. We can fully remove it when SHP is converted to the new style.

test/gtest/shp/shp-tests.cpp
rscohn2: This PR fixes the issue. I will remove the enumerate comment before merging.

include/dr/shp/zip_view.hpp
```diff
@@ -281,6 +296,8 @@ class zip_view : public rng::view_interface<zip_view<Rs...>> {
         rng::begin(std::get<Is>(views_))...);
   }
 
+  template <typename T> T min_many_impl_(T t) const { return t; }
```
rscohn2, line 299: Looks like you could use std::min with an initializer list. I will try before merging.

test/gtest/include/common-tests.hpp
```diff
@@ -325,4 +353,36 @@ inline std::ostream &operator<<(std::ostream &os,
   return os;
 }
 
+template <typename... Ts>
+inline std::ostream &operator<<(std::ostream &os,
```
rscohn2, line 357: And delete these, too.

test/gtest/include/common-tests.hpp
rscohn2: gtest `EXPECT_EQ` requires that the operands can be comparable with `operator==` and printed with `operator<<`. 
To find these operators with ADL, you put it in `std`. Otherwise, you have to put the operator in multiple namespaces. This is only for the testing code so it will not pollute std for regular use of the library.

PR #197, opened 2023-04-12
enable shp tests

rscohn2:


rscohn2:
@BenBrock @lslusarczyk: I have turned on everything that was disabled because of the context exception. A test that always failed within a few minutes has run for several hours.

PR #209, opened 2023-04-17
fixed dv_segments and enabled mhp::take tests

lslusarczyk:
code was failing because dv_segments was treated by ranges library as a container and was wrapped by ref_view. ref_view takes address by object it references. But dv_segments happens to be a temporary object, so its address was becoming invalid. be deriving from view_base we instruct ranges library to treat dv_segments a view, not warp it by ref_view, but simply copy

### DIFF
--- a/include/dr/mhp/containers/distributed_vector.hpp
+++ b/include/dr/mhp/containers/distributed_vector.hpp
@@ -159,7 +159,8 @@ private:
   std::size_t size_;
 }; // dv_segment
 
-template <typename DV> class dv_segments : public std::span<dv_segment<DV>> {
+template <typename DV>
+class dv_segments : public std::span<dv_segment<DV>>, rng::view_base {
 public:
   dv_segments() {}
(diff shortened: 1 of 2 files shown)


rscohn2:
A million thanks for finding that difficult bug!

In addition to being wrong, the code is convoluted. I will try deleting:

https://github.com/oneapi-src/distributed-ranges/blob/9ac85287eb15a1b74a62d8f998c6711d5ff5a1e7/include/dr/mhp/containers/distributed_vector.hpp#L279

and

https://github.com/oneapi-src/distributed-ranges/blob/9ac85287eb15a1b74a62d8f998c6711d5ff5a1e7/include/dr/mhp/containers/distributed_vector.hpp#L162

and instead make segments() return an all_t of segments_
https://github.com/oneapi-src/distributed-ranges/blob/9ac85287eb15a1b74a62d8f998c6711d5ff5a1e7/include/dr/mhp/containers/distributed_vector.hpp#L258

PR #215, opened 2023-04-18
UTs for view::take, split mhp-tests to separate compilation units

lslusarczyk:
Separation speeds up recompilation of just one test suite, which is a common scenario when writing tests. Removed MINIMAL_TEST as I think it is not needed now.


lslusarczyk:
@rscohn2 , please check if you like change I made in mhp-tests compilation, by separating compilation of test suites between .cpp files. If you like I will commit, if not I may revert.

rscohn2:
> @rscohn2 , please check if you like change I made in mhp-tests compilation, by separating compilation of test suites between .cpp files. If you like I will commit, if not I may revert.

Looks good. Thanks for doing this.

rscohn2 (approved):

rscohn2:
> Separation speeds up recompilation of just one test suite, which is a common scenario when writing tests.

With your change: 13 seconds to rebuild one test file, 1 minute to rebuild everything after a header file change (5 minutes CPU time). That's reasonable. I can temporarily comment out tests in the CMakeLists.txt if I want to speed up the build after a header file change.

PR #221, opened 2023-04-20
common tests files placement refactor, moved shp.hpp up

lslusarczyk:
also fixed fields initialization order in dense_matrix_view
and added missing shp/examples to cmake


include/dr/shp.hpp
```diff
@@ -4,10 +4,6 @@
 
 #pragma once
 
-#include <dr/detail/iterator_adaptor.hpp>
```
lslusarczyk, line 7: @BenBrock , please check this change. If something more should be removed or something was wrongly removed, then please write. Files here should include only direct API used by external users (no details), see this thread: https://github.com/oneapi-src/distributed-ranges/pull/190#pullrequestreview-1391885986

rscohn2: I probably added that before we agreed on current approach.

rscohn2 (approved):
Looks very good. Thanks!

include/dr/shp.hpp
lslusarczyk: I will revert this change if we will not agree on shrinking mhp.hpp / shp.hpp by removing non-public headers.

rscohn2:
It would be good to get this merged to avoid conflicts. If the shp examples are not easy to fix, just leave them disabled.

lslusarczyk:
> It would be good to get this merged to avoid conflicts. If the shp examples are not easy to fix, just leave them disabled.

This is true. I've disabled them from running (some of them are meant to be run on GPU only I think), but I want still want them to be compiling in CI.

PR #224, opened 2023-04-20
Implement `gather_test.cpp`

BenBrock:


examples/shp/gather_test.cpp
```diff
@@ -0,0 +1,128 @@
+// SPDX-FileCopyrightText: Intel Corporation
```
lslusarczyk, line 1: Please add compiling of new binaries to cmake. They don't need to be executed, but let's keep CI checking its compilation at least.

scripts/dr-style.py
```diff
@@ -72,9 +72,9 @@
     (r'namespace sycl = ', 'do not define sycl namespace'),
     (r'[< (]cl::sycl', 'delete cl::'),
     (r'parallel_for<class', 'use unnamed lambda'),
-    (r']\(sycl::id<1>', 'use auto instead of id<1> for argument'),
+    # (r']\(sycl::id<1>', 'use auto instead of id<1> for argument'),
```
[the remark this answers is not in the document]
BenBrock: I chatted with Robert and we agreed we could remove these too.

There are cases (such as `__detail::parallel_for` added in this PR) where you want to explicitly specify the types to make it clear you only handle a certain number of dimensions.  In other cases `auto` is probably preferable.

include/dr/shp.hpp
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <dr/detail/ranges_shim.hpp>
```
lslusarczyk, line 7: `details` and `util` are getting back to this file :) can we keep them included in some second-order includes rather than in main shp.hpp?

unless they are really meant to be included by user and used directly by user

scripts/dr-style.py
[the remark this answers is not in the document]
lslusarczyk: OK. Thank you for discussing it with Robert. So let's remove these rules (not keep them as commented code).

include/dr/shp.hpp
BenBrock: Fair point, will remove.

PR #225, opened 2023-04-24
zip_view with base + support for distributed/remote ranges

rscohn2:
WIP

### DIFF
--- a/test/gtest/mhp/zip-base.cpp
+++ b/test/gtest/mhp/zip-base.cpp
@@ -359,7 +359,6 @@ TEST(Zip, CopyConstructor) {
   EXPECT_EQ(dist, dist_copy);
 }
 
-#if 0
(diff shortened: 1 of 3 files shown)


test/gtest/mhp/zip-base.cpp
```diff
@@ -0,0 +1,405 @@
+// SPDX-FileCopyrightText: Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "xhp-tests.hpp"
+
+template <typename... Rs> auto test_zip(Rs &&...rs) {
+  return dr::mhp::views::zip_base(std::forward<Rs>(rs)...);
```
rscohn2, line 8: @suyashbakshi: `zip_base` is a zip that should be able to handle dot product. Can you try it?

PR #230, opened 2023-04-24
mhp switch local to use segment, not iterator

rscohn2:
MHP constructed a local range from local iterators. Switch to requesting a local range from the segment. MHP always has a segment when it needs local iterators and this will simplify some views. The CPO for getting the local range will construct it when only the iterators implement local.

The CPO used `std::span` to construct the local range, but the constructor it used only works for contiguous range and some views were not providing that. Switch to `ranges::views::counted``.

### DIFF
--- a/include/dr/detail/ranges.hpp
+++ b/include/dr/detail/ranges.hpp
@@ -152,8 +152,10 @@ struct local_fn_ {
              std::contiguous_iterator<rng::iterator_t<R>> ||
              rng::contiguous_range<R>)
   auto operator()(R &&r) const {
-    if constexpr (has_local_method<rng::iterator_t<R>>) {
-      return std::span(rng::begin(r).local(), rng::size(r));
+    if constexpr (has_local_method<R>) {
+      return r.local();
+    } else if constexpr (has_local_method<rng::iterator_t<R>>) {
+      return rng::views::counted(rng::begin(r).local(), rng::size(r));
     } else if constexpr (has_local_adl<R>) {
       return local_(std::forward<R>(r));
     } else if constexpr (std::contiguous_iterator<rng::iterator_t<R>>) {
--- a/include/dr/mhp/views/views.hpp
+++ b/include/dr/mhp/views/views.hpp
@@ -16,8 +16,7 @@ template <typename R> auto local_segments(R &&dr) {
   };
   // Convert from remote iter to local iter
   auto local_iter = [](const auto &segment) {
-    auto b = dr::ranges::local(rng::begin(segment));
-    return rng::subrange(b, b + rng::distance(segment));
+    return dr::ranges::local(segment);
   };
   return dr::ranges::segments(std::forward<R>(dr)) |
          rng::views::filter(is_local) | rng::views::transform(local_iter);


include/dr/detail/ranges.hpp
```diff
@@ -152,8 +152,10 @@ struct local_fn_ {
              std::contiguous_iterator<rng::iterator_t<R>> ||
              rng::contiguous_range<R>)
   auto operator()(R &&r) const {
-    if constexpr (has_local_method<rng::iterator_t<R>>) {
-      return std::span(rng::begin(r).local(), rng::size(r));
+    if constexpr (has_local_method<R>) {
```
lslusarczyk, line 155: this operator is having 5 requires, but 4 _if_ branches. rng::contiguous_range<R> is missing its _if_

I think pattern with separated operators is more readable than repeated _if constexpr_ and also immune for missing _if_ bug.
```
template <class R> requires(rng::forward_range<R> && has_local_adl<R>)
auto operator()(R &&r) const { return r.local(); }

template <class R> requires(rng::forward_range<R> && has_local_method<rng::iterator_t<R>>)
auto operator()(R &&r) const { return ...; }

template <rng::contiguous_range<R>>
auto operator()(R &&r) const { return ...;}

...
```
This is the pattern I see in ranges-v3 library - many operator() methods without _if_-s inside. E.g. in [all view](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/all.hpp#L70) or [cartesian_product view](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/cartesian_product.hpp#L461)

include/dr/detail/ranges.hpp
```diff
@@ -152,8 +152,10 @@ struct local_fn_ {
...
+      return r.local();
+    } else if constexpr (has_local_method<rng::iterator_t<R>>) {
+      return rng::views::counted(rng::begin(r).local(), rng::size(r));
     } else if constexpr (has_local_adl<R>) {
       return local_(std::forward<R>(r));
     } else if constexpr (std::contiguous_iterator<rng::iterator_t<R>>) {
```
lslusarczyk, line 161: If we prefer `rng::views::counted` over `std::span` let's change also this second place.

I understand motivation here is to not use views from std, but from ranges library, as views from std are not compatible with views from ranges-v3.

But I think `subrange(begin,end)` may be better - it should not need to iterate twice over `forward_range` to first get its size and then find its end (see code [here](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/counted.hpp#L104)).

rscohn2: counted is a subrange for random access iterators and it also handles non random access.

include/dr/detail/ranges.hpp
lslusarczyk: We may always make them mutually exclusive. I understand compiler will catch such a cases and force programmer to think what he/she wants instead of silently rely on order inside `if ... else if ... else ...`.

But if we don't have strong opinion/agreement about what is definitely better let's don't do such refactors.

include/dr/detail/ranges.hpp
lslusarczyk: Can [subrange(Iterator && i, Sentinel s)](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/view/subrange.hpp#L211) handle non-random access? Looks like it can. I've not checked it. If so, then it is more efficient.

PR #232, opened 2023-04-25
use slice instead of remote_subrange

rscohn2:
Use rng::slice_view instead of remote_range. It works for mhp, but does not compile with shp. I think the problem is that segment_tools uses rng::zip_view and it does not propagate a rank.

Setting this aside.


rscohn2:
Not pursuing this. Would need to support `dr::ranges::local` in all the segments.

PR #237, opened 2023-04-27
Update `gemv` to use `duplicated_vector`

BenBrock:


BenBrock:
For now, we are allowing users to explicitly handling the management of temporary memory needed for `gemv`.

```cpp
// This is temporary memory where `b` will be gathered on each device.
shp::duplicated_vector<float> b_duplicated(k);

shp::gemv(c, a, b, b_duplicated);
```

This is necessary because memory allocation is expensive, and in iterative algorithms we don't want to do it in every iteration.  I would like to obviate this need by setting up a fast memory allocator.  It would call `malloc_device` when needed, but then hold on to the memory to re-use in subsequent allocations.

In particular, matrix multiplication is a lot simpler with a fast and efficient temporary memory allocator.

PR #245, opened 2023-05-05
new mhp zip view

rscohn2:
Only relies on `zip_view` and iterator having a base field.  It should be possible to implement the support for distributed with CPO's, but for now, local/segments/rank are MHP-specific methods. 

This implementation eliminates some of the issues with the earlier implementation. The views/iterators do not need to store std::vector. Instead, it implements the segments as a forward_range instead of a random_access_range. The iterator does not have pointers to the view.

### DIFF
--- a/test/gtest/common/zip.cpp
+++ b/test/gtest/common/zip.cpp
@@ -4,6 +4,10 @@
 
 #include "xhp-tests.hpp"
 
+template <typename... Rs> auto test_zip(Rs &&...rs) {
+  return xhp::views::zip(std::forward<Rs>(rs)...);
+}
+
 // Fixture
 template <typename T> class Zip : public testing::Test {
 public:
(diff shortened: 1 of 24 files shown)


test/gtest/common/zip.cpp
```diff
@@ -4,176 +4,206 @@
...
-  EXPECT_EQ(rops.vec1, xops.vec1);
+  auto local = rng::views::zip(ops.vec);
+  auto dist = test_zip(ops.dist_vec);
+  auto flat = rng::views::join(dr::ranges::segments(dist));
+  EXPECT_TRUE(is_equal(local, flat));
 }
 
-TYPED_TEST(Zip, Dist1) {
+#if 0
+// Will not compile with SHP
```
rscohn2, line 63: @BenBrock. This is a new test that I am disabling because it will not compile with SHP. It looks like SHP zip_view does provide segments from an iterator. It is not urgent, but would be good to fix.

PR #253, opened 2023-05-10
mhp: copy from/to local/remote contiguous ranges

rscohn2:
also change dot_product_benchmark to initialize with random numbers on root and copy to other processes

### DIFF
--- a/examples/mhp/dot_product_benchmark.cpp
+++ b/examples/mhp/dot_product_benchmark.cpp
@@ -17,7 +17,7 @@
 
 namespace mhp = dr::mhp;
 
-using T = float;
+using T = double;
 
 MPI_Comm comm;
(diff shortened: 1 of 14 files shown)


examples/mhp/dot_product_benchmark.cpp
```diff
@@ -164,34 +164,30 @@ int main(int argc, char **argv) {
...
+    std::cout << options_spec.help() << "\n";
+    exit(0);
+  }
+
   n = options["n"].as<std::size_t>();
   n_iterations = options["i"].as<std::size_t>();
   std::vector<T> x_local(n);
   std::vector<T> y_local(n);
-#if 0
+
```
rscohn2, line 176: @suyashbakshi: changed over to random initialization

PR #268, opened 2023-05-17
drop testing of float vectors

rscohn2:
CI takes 30 minutes, drop testing of float vectors and only test int to see if that will speed it up.

### DIFF
--- a/test/gtest/mhp/xhp-tests.hpp
+++ b/test/gtest/mhp/xhp-tests.hpp
@@ -34,17 +34,8 @@ inline void fence() { dr::mhp::fence(); }
 
 #include "common-tests.hpp"
 
-#ifdef QUICK_TEST
-
 // minimal testing for quick builds
 using AllTypes = ::testing::Types<dr::mhp::distributed_vector<int>>;
-
-#else
-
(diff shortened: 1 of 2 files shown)


BenBrock:
Are we still testing equality of a distributed range and a local range with `dv == v`?  If so, we should probably replace that with something like:

```cpp
template <typename T, typename U>
void assert_equals(const T& t, const U& u) {
  // Not sure exactly what the Gtest macro is
  EXPECT_TRUE(t == u);
}

template <std::floating_point T, std::floating_point U>
void assert_equals(T t, U u) {
  // Do something smart for floating point
}

template <rng::forward_range R1, rng::forward_range R2>
void assert_equals(R1&& r1, R2&& r2) {
  for (auto&& [a, b] : xhp::views::zip(r1, r2)) {
    assert_equals(a, b);
  }
}

template <rng::distributed_range R1, rng::forward_range R2>
void assert_equals(R1&& r1, R2&& r2) {
  using T = rng::range_value_t<R1>;
  std::vector<T> v(rng::distance(r1));

  xhp::copy(rng::begin(r1), rng::end(r1), v.data());

  assert_equals(v, r2);
}
```

PR #276, opened 2023-05-20
gbench for shp and some mhp fixes

rscohn2:


benchmarks/gbench/common/distributed_vector.cpp
```diff
@@ -39,6 +39,9 @@ static void Fill_Local(benchmark::State &state) {
 
 BENCHMARK(Fill_Local);
 
+#ifndef BENCH_SHP
+// Not defined?
```
rscohn2, line 43: @BenBrock Not sure why this won't compile with SHP

benchmarks/gbench/common/distributed_vector.cpp
```diff
@@ -126,6 +130,9 @@ static void TransformIdentity_Local(benchmark::State &state) {
 
 BENCHMARK(TransformIdentity_Local);
 
+#ifndef BENCH_SHP
+// segfault
```
rscohn2, line 134: @BenBrock SHP segfaults here

PR #278, opened 2023-05-22
slide view for 1D

lslusarczyk:


rscohn2 (commented):
The sliding view looks ok in general, but I want to understand the changes to the implementation of local after you give some more detail.

examples/mhp/stencil-1d.cpp
```diff
@@ -7,27 +7,19 @@
...
   rng::fill(b, 0);
 
-  auto in = rng::subrange(a.begin() + 1, a.end() - 1);
-  auto out = rng::subrange(b.begin() + 1, b.end() - 1);
+  auto in = rng::ref_view(a);
+  auto out = rng::ref_view(b);
+
   for (std::size_t s = 0; s < steps; s++) {
-    rng::transform(in, out.begin(), stencil_op);
+    rng::transform(rng::sliding_view(in, 3), out.begin() + 1, stencil_op);
```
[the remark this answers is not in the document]
lslusarczyk: Good point. Applied.

include/dr/detail/logger.hpp
```diff
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include <fmt/core.h>
```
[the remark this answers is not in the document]
lslusarczyk: right, applied

include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -44,22 +44,27 @@ template <typename DV> class dv_segment_iterator {
 
   // Comparison
   bool operator==(const dv_segment_iterator &other) const noexcept {
+    assert(dv_ != nullptr && dv_ == other.dv_);
```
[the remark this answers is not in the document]
lslusarczyk: You are right. E.g. `rng::join_view` creates iter by default constructor and compares it to other default constructed iter. `-D_GLIBCXX_DEBUG` also allows to compare two singular iterators. I've precised assertion conditions.

> I use dv_ to indicate an unreachable sentinel.

Do you mean that when `dv_` is a `nullptr` then it is an unreachable sentinl, right?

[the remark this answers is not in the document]
rscohn2: > Do you mean that when `dv_` is a `nullptr` then it is an unreachable sentinl, right?

Right.

test/gtest/mhp/slide_view.cpp
```diff
@@ -0,0 +1,52 @@
...
+TYPED_TEST_SUITE(Slide, AllTypes);
+
+TYPED_TEST(Slide, is_compliant) {
+  TypeParam dv(10, dr::halo_bounds(2, 1, false));
+  LocalVec<TypeParam> lv(10);
+  iota(dv, 100);
+  std::iota(rng::begin(lv), rng::end(lv), 100);
+
+  auto local_sliding_view = rng::sliding_view(lv, 4); // halo_bounds + 1
+  auto dv_sliding_view [[maybe_unused]] = xhp::views::sliding(dv);
```
rscohn2, line 19: I think there is a benefit to supporting a sliding api that takes a window size, even if passing anything other than the halo size is an error. My main motivation is that the view of the same name in ranges-v3 takes a window size. Secondary is for readability. I don't object to making the window size optional.

include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -127,12 +142,46 @@ template <typename DV> class dv_segment_iterator {
...
   void put(const value_type &value) const { put(&value, 1); }
 
-  auto rank() const { return segment_index_; }
-  auto local() const { return dv_->data_ + index_ + dv_->halo_bounds_.prev; }
+  auto rank() const {
+    assert(dv_ != nullptr);
+    return segment_index_;
+  }
+
+  auto local() const {
```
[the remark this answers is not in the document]
lslusarczyk: Consider `DV<int>`, `halo.radious=1`, rank 5 having elements pointed by dv_segment_iterators `(segment_index_:5,index_:0)`, `(segment_index_:5,index_:1)`, `(segment_index_:5,index_:2)`.

sliding_view at rank 5 consists of three subranges which will be converted to local ones:

1. from `(segment_index_:4,index_:2)` to `(segment_index_:5,index_:1)`
2. from `(segment_index_:5,index_:0)` to `(segment_index_:5,index_:2)`
3. from `(segment_index_:5,index_:1)` to `(segment_index_:6,index_:0)`

All of these iterator needs to be converted into `int*` at rank 5. Thus we need to be able to convert some iterators having segment_index_ 4 and 6 also on rank 5 using halo if halo covers such iterators.

include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -296,9 +356,19 @@ class distributed_vector {
...
   auto &halo() { return *halo_; }
+  auto &halo_bounds() const { return halo_bounds_; }
 
   auto segments() const { return rng::views::all(segments_); }
 
+  void dump_to_log(std::string desc = "") const {
+    dr::drlog.debug("DV {}, dataAddr:{}", desc,
+                    static_cast<void *>(this->data_));
+    for (std::size_t idx = 0; idx < data_size_; ++idx)
+      dr::drlog.debug(" - idx:{} val:{} addr:{}", idx, this->data_[idx],
```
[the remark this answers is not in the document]
lslusarczyk: right, without value it gets useless, removed whole `dump_to_log`. It needs to be done in better way in other PR with printing values if possible.

include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -127,12 +142,46 @@ template <typename DV> class dv_segment_iterator {
...
+  auto local() const {
+    assert(dv_ != nullptr);
+    const auto my_process_segment_index = dv_->win_.communicator().rank();
+
+    if (my_process_segment_index == segment_index_)
+      return dv_->data_ + index_ + dv_->halo_bounds_.prev;
+
+    assert(!dv_->halo_bounds().periodic); // not implemented
+
+    if (my_process_segment_index + 1 == segment_index_) {
```
rscohn2, line 164:
```suggestion
    // Sliding view needs local iterators that point to the halo
    
    if (my_process_segment_index + 1 == segment_index_) {
```

test/gtest/mhp/slide_view.cpp
lslusarczyk: If we are going to support ranges-v3 like API, then we need to assume where the center is (center determines which rank given subrange belongs to).

`sliding_view(dr, 5)` -> center seems natural to be the 3th element
`sliding_view(dr,6)` -> center is the 3rd or 4th element?

What about case when `halo.prev != halo.next`, 
e.g. halo.prev=1, halo.next=10
where should be center of `sliding_view(dr, 5)`? natural 3rd element is incorrect

We could let user to specify center but then we end up with incompatible API - same situation as now.

We can remove `prev!=next` halo (halo can have one radius only)

We can rename sliding_view to halo_view.

These were my dilemmas. No solution seems to be perfect so I selected the simplest one. What do you propose?

examples/mhp/stencil-1d.cpp
```diff
@@ -50,14 +42,19 @@ int stencil(auto n, auto steps) {
...
 
-  auto in = rng::subrange(a.begin() + 1, a.end() - 1);
-  auto out = rng::subrange(b.begin() + 1, b.end() - 1);
+  auto in = rng::ref_view(a);
+  auto out = rng::ref_view(b);
+
   for (std::size_t s = 0; s < steps; s++) {
-    dr::mhp::halo(in).exchange();
-    dr::mhp::transform(in, out.begin(), stencil_op);
+    dr::mhp::halo(in.base()).exchange();
```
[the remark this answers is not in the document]
lslusarczyk: `base()` is already not needed in this code - removed

Making `dr::mhp::halo(dr::mhp::views::sliding(in)).exchange()` is not obvious how to me how to do it in the cleanest way now. I've left a TODO. If you think this is important I'll create task for it or I'll think and propose some solution in this PR. 

Is it important to do / important to do before merging this PR?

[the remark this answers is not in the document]
rscohn2: Sliding is for stencil and stencil needs halo exchange. But it can be a separate commit. We really need a universal way to get a handle to the underlying container that dr::mhp::halo() and other CPO's can use, instead of trying to handle each CPO individually.

lslusarczyk:
All comments applied. But `icpx` build shows that even empty `xhp::for_each(xhp::views::sliding(dv), [](auto &&r) {})` does not compile with `-fsycl` flag.
```
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/test/gtest/mhp/slide_view.cpp:5:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/test/gtest/mhp/xhp-tests.hpp:7:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/include/./dr/mhp.hpp:41:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/include/./dr/detail/ranges_shim.hpp:21:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/all.hpp:17:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/action.hpp:17:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/action/action.hpp:23:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/action/concepts.hpp:23:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/range/concepts.hpp:41:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/range/access.hpp:37:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/iterator/reverse_iterator.hpp:20:
In file included from /home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/iterator/basic_iterator.hpp:30:
/home/lucasso/dds/libraries.runtimes.hpc.dds.distributed-ranges/cmake-build-oneapi/_deps/range-v3-src/include/range/v3/utility/box.hpp:308:20: error: SYCL kernel cannot use a non-const global variable
```
I'll analyze it tomorrow.

lslusarczyk:
@rscohn2 , could you please have a second look? I've applied all your comments and added support for transform.
Most important changes since last time you reviewed are

1. changes in zip.hpp 
2. last two test cases in slide_view.cpp
3. fsycl flag added to stencil-1d binary, which (together with above mentioned 2 test cases) ensure sliding view works now in sycl

include/dr/mhp/views/views.hpp
```diff
@@ -16,11 +16,18 @@ template <typename R> auto local_segments(R &&dr) {
   auto is_local = [](const auto &segment) {
     return dr::ranges::rank(segment) == default_comm().rank();
   };
+
   // Convert from remote iter to local iter
   auto local_iter = [](const auto &segment) {
-    auto b = dr::ranges::local(rng::begin(segment));
-    return rng::subrange(b, b + rng::distance(segment));
+    if constexpr (localizable_range<decltype(*rng::begin(segment))>) {
```
rscohn2, line 22: Can you explain what is happening here?

include/dr/mhp/views/zip.hpp
```diff
@@ -83,6 +83,31 @@ template <typename Base> auto base_to_segments(Base &&base) {
   return std::apply(bases_to_segments, base);
 }
 
+// based on https://ericniebler.github.io/range-v3/#autotoc_md30  "Create custom
+// iterators"
+template <typename Iter> struct cursor_over_local_ranges {
```
rscohn2, line 88: It does look like a nice way to define an iterator. We would have to eliminate the range-v3 dependency.

include/dr/mhp/views/zip.hpp
```diff
@@ -193,11 +218,17 @@ class zip_iterator {
...
 private:
-  // If it is not a remote iterator, assume it is a local iterator
-  auto static base_local(auto iter) { return iter; }
-
-  auto static base_local(dr::remote_iterator auto iter) {
-    return dr::ranges::local(iter);
+  auto static base_local(auto iter) {
+    if constexpr (dr::remote_iterator<decltype(iter)>) {
+      return dr::ranges::local(iter);
+    } else if constexpr (dr::localizable_range<decltype(*iter)>) {
```
rscohn2, line 224: Still confused by all this.

include/dr/mhp/containers/distribution.hpp
```diff
@@ -11,7 +11,19 @@ namespace dr::mhp {
...
+    return *this;
+  }
+
+  distribution &halo(std::size_t prev, std::size_t next) {
+    halo_bounds_.prev = prev;
+    halo_bounds_.next = next;
+    return *this;
+  }
+
+  distribution &periodic() {
```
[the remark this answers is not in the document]
lslusarczyk: I see. I'm fixing.

include/dr/mhp/views/views.hpp
lslusarczyk: This _if_ checks if segment consists of ranges which can be made local. This is the case of sliding_view which consists of not elements (that can be made local) but ranges of elements (that can be made local).

Previously there was _is_sliding_view_ here, but IMO the current form is better as it is more general.

include/dr/mhp/views/zip.hpp
lslusarczyk: I may copy basic_iterator from ranges-v3 to our code base (change its namespace, remove not needed code from it, use it here). What do you think?

rscohn2: It would be ok if you could isolate just this one thing without pulling the rest of ranges-v3

test/gtest/mhp/slide_view-3.cpp
rscohn2, line 1: Can you do tests:
* zip 2 sliding views together and pass to a for_each
* transform_view a sliding view and pass for a for_each

include/dr/concepts/concepts.hpp
```diff
@@ -56,4 +56,9 @@ concept distributed_contiguous_iterator =
     remote_contiguous_range<rng::range_value_t<decltype(dr::ranges::segments(
         std::declval<Iter>()))>>;
 
+template <typename R>
+concept localizable_range = rng::range<R> && requires(R &r) {
+  { dr::ranges::local(rng::begin(r)) } -> std::contiguous_iterator;
```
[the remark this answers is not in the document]
lslusarczyk: It is not used to describe properties of iterator which traverses over zip, but to describe properties of sub-iterator which zip uses, that is an iterator which traverses over one of zipped collections.

I need to detect a range which has no rank() CPO working on it, but has local() CPO working on it.

I've just wrote it in non-confusing way I think - please see: https://github.com/oneapi-src/distributed-ranges/pull/278/commits/dacd410da4a86f36afe49d5bd73e57487e995303#diff-0f733b699e6a6036d457bfc38247e734fbe98c05e6c0b40a2072310009cb77f8

The old remote_contiguous_range is decomposed to be remote_range + localizable_contiguous_range

**Why we need such a concept.**

_segments_ CPO produces the following segments:
```
dr::remote_subrange<
|  ranges::basic_iterator<
|  |  ranges::adaptor_cursor<
|  |  |  dr::mhp::distributed_vector<int>::iterator,
|  |  |  ranges::sliding_view<
|  |  |  |  ranges::ref_view<
|  |  |  |  |  dr::mhp::distributed_vector<int>>,
|  |  |  |  ranges::sliding_view_detail::cache::none>
|  |  |  ::adaptor<true>>>>
```
Such a segment has `rank()`. It consists of a range (let's name it a _big range_) of ranges (let's name them _small ranges_). Big range has an iterator pointing to small range. Small range has iterator (this case: `mhp::distributed_vector::iterators`) which  has `local()`. `local()` does not work on _big range_ iterators. 

If we can handle situation: iterator points to range which has local iterator, then we avoid implementing sliding_view (and possibly more views producing range of ranges) on our own. Thanks to this I just needed to add _segments_ CPO for ranges::sliding_view.

Similar situation is for _zip_view_. Type of _iter_ param passed to _base_local()_ is 
```
ranges::basic_iterator<
|  ranges::adaptor_cursor<
|  |  dr::mhp::distributed_vector<int>::iterator,
|  |  ranges::sliding_view<
|  |  |  ranges::ref_view<
|  |  |  |  dr::mhp::distributed_vector<int>>,
|  |  |  ranges::sliding_view_detail::cache::none>
|  |  ::adaptor<true>>>
```
Again, it is an iterator which has no `local()` working on it, but points to a range which has. I need to handle this range of local ranges by iterator wrapper I wrote and detect it by concept I added.

Next step would be to handle/detect in generic way _range of ... range of .. (times N) .. or ranges which can be made local_, which I expect will handle for free all tricky views (like sliding_view<sliding_view<...>>`) e.g. for stencil  2D.

include/dr/mhp/views/zip.hpp
lslusarczyk: I tried to explain it in the answer of previous comment (see: _Not sure what this means. local zip interator is not contiguous..._) and refactored concept in https://github.com/oneapi-src/distributed-ranges/pull/278/commits/dacd410da4a86f36afe49d5bd73e57487e995303 to make it fitting into all of our concepts.

test/gtest/mhp/slide_view-3.cpp
lslusarczyk: sure, I will add, should be valuable, thank you for this idea

test/gtest/mhp/slide_view.cpp
```diff
@@ -124,4 +124,73 @@ TYPED_TEST(Slide, slide_works_with_transform) {
...
+
+  EXPECT_EQ(11 + 12, dv_2[0]);
+  EXPECT_EQ(12 + 13, dv_2[1]);
+  EXPECT_EQ(13 + 14, dv_2[2]);
+  EXPECT_EQ(14 + 15, dv_2[3]);
+  EXPECT_EQ(24, dv_2[4]);
+  EXPECT_EQ(25, dv_2[5]);
+}
+
+// TODO: slide doesn't work on transformed range, working on it
```
rscohn2, line 167: This is good to test, but I was asking for `slide | transform` similar to `two_slides_can_be_zipped_and_read_by_foreach`

include/dr/detail/ranges.hpp
```diff
@@ -140,7 +140,8 @@ template <iter_has_local_method T>
 struct is_localizable_helper<T> : std::true_type {};
 
 template <rng::forward_iterator Iter>
-  requires(not iter_has_local_method<Iter> && not has_local_adl<Iter>)
+  requires(not iter_has_local_method<Iter> && not has_local_adl<Iter>) &&
+          requires() { typename Iter::value_type; }
```
rscohn2, line 144:
```suggestion
          requires() { std::iter_value_t<Iter>; }
```
Did not check that this works

examples/mhp/stencil-1d.cpp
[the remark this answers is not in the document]
lslusarczyk: I can. However I think old way of writing stencil becomes outdated when we have `sliding_view`. 
```
auto stencil_op = [](auto &&v) {
  auto p = &v;
  return p[-1] + p[0] + p[+1];
};
```
It has one advantage of using 0 and negative indices.
It requires hack with taking address of value, it does not check for boundaries and for halo region. It is not range-style.
I would advertise it to be used no more, so I changed example instead of adding one.

If you disagree I will add previous version back. Please decide.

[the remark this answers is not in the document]
rscohn2: The new one is preferred, but the current one is more flexible. It is ok to make the new one stencil-1d.cpp and rename the current one to: stencil-1d-pointer.cpp, but please do not delete it. Another thing to check is stencil-1d-array to check whether sliding supports elements that are not simple int/float.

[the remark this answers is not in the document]
lslusarczyk: > ... rename the current one to: stencil-1d-pointer.cpp ...

done

> ... to check is stencil-1d-array to check whether sliding supports elements that are not simple int/float

It looks like a code that is better to be placed in a test, not example. I've added such a test.

include/dr/detail/ranges.hpp
lslusarczyk: applied, works

include/dr/mhp/containers/distributed_vector.hpp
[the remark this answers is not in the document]
lslusarczyk: Let me add an iterator factory (and change iterator I defined, that uses basic_iterator) in a separate PR after we merge this changes if you're OK with them.

lslusarczyk:
@rscohn2 , I think I applied everything. If changes are OK please merge.

include/dr/mhp/views/segmented.hpp
```diff
@@ -69,6 +71,9 @@ class segmented_view : public rng::view_base {
                                    rng::end(segments_tpl_));
   }
 
+  auto size() const { return rng::size(segments_tpl_); }
```
rscohn2, line 74: Inheriting from [view_interface](https://en.cppreference.com/w/cpp/ranges/view_interface) instead of `view_base` is preferred as it will define `size()` and `empty()`. One of us can do that in a separate commit.

https://github.com/oneapi-src/distributed-ranges/blob/3beb3f522cb7f28a0c64926ac782f695a80fbed5/include/dr/shp/zip_view.hpp#L108

PR #292, opened 2023-05-28
switch mhp to use local on remote ranges instead of iterators

rscohn2:
Draft because still experimenting


rscohn2:
Closing, Not pursuing.

PR #294, opened 2023-05-29
Added the BenchmarkRunner class  and SYCL dp benchmarks

haichangsi:
Added a new python class to run benchmarks faster and easier.

Added 2 new dot product benchmarks based on SYCL only - using USM and buffers.

Introduced CsvPlotter to autogenerate plots.


rscohn2:
I added google benchmark. There is dot product, fill, copy, stencil, ... It is a common infrastructure for running/reporting multiple benchmarks. It supports shp/mhp as well as using DPL/SYCL with DR.

Can you add whatever you need there instead?

examples/shp/dot_product_benchmark.cpp
```diff
@@ -160,26 +197,31 @@ int main(int argc, char **argv) {
...
-  fmt::print("Result: {}\n", sum);
 #endif
 
+  //  fmt::print("----------------------------------------------------------\n");
+  //  fmt::print("devices, size, seq_time, shp_time, dpl_time, mkl_time\n");
+  //  fmt::print("{}, {}, {}, {}, {}, {}\n", devices.size(), size, seq_median,
+  //  shp_median, dpl_median, mkl_median);
+
+  // fmt::print("----------------------------------------------------------\n");
+  // fmt::print("devices, size, shp_time, dpl_time\n");
```
mateuszpn, line 223: The code that is no longer used should be removed, especially if it is trivial (like printing messages).

examples/shp/dot_product_benchmark.cpp
```diff
@@ -50,23 +50,44 @@ auto dot_product_sequential(X &&x, Y &&y) {
 }
 
 int main(int argc, char **argv) {
+
+  std::size_t size = 0; // size of problem
+  std::size_t ndev = 0; // size of problem
```
mateuszpn, line 55: "// number of devices" probably. Comments are good if they are precise

examples/shp/dot_product_sycl.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  std::vector<double> durations;
+  durations.reserve(dev_num);
+
+  for (std::size_t i = 0; i < devices.size(); i++) {
+    auto begin = std::chrono::high_resolution_clock::now();
+
+    queues.emplace_back(
+        sycl::queue(devices[i], sycl::property::queue::in_order{}));
+    std::size_t chunk_size = vec_size / dev_num;
```
mateuszpn, line 72: You get better (more uniform) distribution when last node is less loaded than the others:
chunk_size = vec_size / (dev_num - 1);
last_chunk_size = vec_size - chunk_size * (dev_num - 1);

examples/shp/dot_product_sycl_buffers.cpp
```diff
@@ -0,0 +1,135 @@
...
+int main(int argc, char **argv) {
+
+  if (argc != 3) {
+    return 1;
+  }
+
+  std::size_t dev_num = std::atoi(argv[1]);
+  std::size_t vec_size = std::atoi(argv[2]);
+
+  if (vec_size == 0 || dev_num == 0) {
```
mateuszpn, line 31: Consider a message here, do not finish with no word.

mateuszpn (changes requested):
In general, I'd give a gbench (mentioned by Robert) a look, and probably make benchmarks uniform. The system is similar to gtest. The extra work will bring time savings in the future (we will do a lot of benchmarks).

rscohn2:
> It should be closed probably unless we want to use a part of it (pure SYCL dot product code @lslusarczyk ) @rscohn2

Closing, we can refer to the PR if we need the code.

PR #300, opened 2023-06-01
tests for dr::views::iota

mateuszpn:


test/gtest/mhp/iota.cpp
rscohn2, line 1: Move from mhp to common

test/gtest/mhp/iota.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  EXPECT_TRUE(
+      equal(dv, std::vector<int>{-1, -2, -3, -4, -5, -6, -7, -8, -9, -10}));
+}
+
+TYPED_TEST(IotaView, ReduceRootRange) {
+  auto v = dr::views::iota(1, 11);
+  auto root = 0;
+
+  auto result = dr::mhp::reduce(root, v, 0, std::plus{});
```
rscohn2:
```suggestion
  auto result = xhp::reduce(v, 0, std::plus{});
```

In MHP, all ranks will get the result of the reduce if you don't pass a root. This makes it possible to share the test.

test/gtest/common/iota_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+TYPED_TEST_SUITE(IotaView, AllTypes);
+
+TYPED_TEST(IotaView, ZipWithDR) {
+  xhp::distributed_vector<int> dv(20);
+  auto v = dr::views::iota(1, 20);
+  int ref = 1;
+
+  auto z = xhp::views::zip(dv, v);
+  for (auto itr = z.begin(); itr != z.end(); itr++) {
```
rscohn2, line 19:
```suggestion
for (auto v : z) {
```

test/gtest/common/iota_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+    EXPECT_EQ(dv[ref - 1], ref);
+    ref++;
+  }
+}
+
+TYPED_TEST(IotaView, Copy) {
+  TypeParam dv(10);
+  auto v = dr::views::iota(1, 11);
+
+  rng::copy(v, dv.begin());
```
rscohn2, line 31:
```suggestion
  xhp::copy(v, dv.begin());
```

And same for other algorithms

test/gtest/common/iota_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+TYPED_TEST_SUITE(IotaView, AllTypes);
+
+TYPED_TEST(IotaView, ZipWithDR) {
+  xhp::distributed_vector<int> dv(20);
+  auto v = dr::views::iota(1, 20);
+  int ref = 1;
+
+  auto z = xhp::views::zip(dv, v);
+  for (auto itr = z.begin(); itr != z.end(); itr++) {
+    auto [dve, ve] = *itr;
```
rscohn2, line 20:
```suggestion
    auto [dve, ve] = v;
```

test/gtest/common/iota_view.cpp
mateuszpn: dr:views::iota_view has only rng::views::iota below it
As it is not a view at real data, it has no segments. We can:
- change the requirements for input in algorithms (output will be distributed_iterator anyway); for dr::mhp::transform it is enough, because it has dr::mhp::zip inside, and execution is distributed anyway, not sure what to do with other algorithms
- do a segments() function to meet the distributed_range requirement; as we have no physical storage, what should it return?

rscohn2: I prefer the first option. MHP has a SPMD programming model, and my thinking is that algorithms can assume that arguments that are not distributed_ranges are replicated across the processes and can be segmented as desired. Zip does that, but I only tested for iota. This is not what you want for scatter/gather where it is an all-to-1 or 1-to-all communication (e.g. distribute a local vector resident on a single rank). For that case, I have overloads that also accept a root rank.

@BenBrock: I know that SHP zip also handles iota. What is the general model?

mateuszpn: Made copy() specialized for dr::views::iota in [fd4ca86](https://github.com/oneapi-src/distributed-ranges/pull/300/commits/fd4ca869190e32fa75ce9e73edd3a23ad8cc47b6). As iota view is generated, not stored, it required scattering subsequent values from root node to distributed segments.

include/dr/mhp/algorithms/copy.hpp
```diff
@@ -61,4 +62,26 @@ void copy(std::size_t root, rng::contiguous_range auto &&in,
   barrier();
 }
 
+// copy for dr::views::iota
```
rscohn2, line 60: A copy with root for iota is not needed and this implementation would be very slow. It would be better to generalize the constraints on the first argument type here: https://github.com/oneapi-src/distributed-ranges/blob/fd4ca869190e32fa75ce9e73edd3a23ad8cc47b6/include/dr/mhp/algorithms/copy.hpp#L13

lslusarczyk (changes requested):

include/dr/mhp/algorithms/copy.hpp
```diff
@@ -5,11 +5,12 @@
 #pragma once
 
 #include <dr/detail/ranges_shim.hpp>
+#include <dr/views/iota.hpp>
```
lslusarczyk, line 8: not needed

CMakeLists.txt
```diff
@@ -168,7 +168,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     add_compile_options(-Wno-tautological-constant-compare)
     # DPL turns on -fopenmp-simd, triggering optimization warnings in parallel stl
     add_compile_options(-Wno-error=pass-failed)
-    add_compile_options(-Wno-error=unknown-pragmas)
+    add_compile_options(-Wno-unknown-pragmas)
```
lslusarczyk, line 171: revert this change

test/gtest/shp/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_executable(
   ../common/fill.cpp
   ../common/for_each.cpp
   ../common/iota.cpp
+  ../common/iota.cpp
```
lslusarczyk, line 19: double added iota.cpp, remove

test/gtest/mhp/CMakeLists.txt
```diff
@@ -63,6 +64,7 @@ foreach(test-exec IN ITEMS mhp-tests mhp-tests-3 mhp-quick-test)
     cxxopts
     DR::mpi
   )
+
```
lslusarczyk, line 67: Next time please avoid making empty changes to CMakeFiles (like empty spaces or lines). They make review harder.
If they are made by your IDE try to find option to disable them or clear them before making PR.

PR #314, opened 2023-06-06
Fix copy

BenBrock:
Implement distributed -> distributed version of `shp::copy`, as well as range-based interface, and enable shared copy tests for `shp`.

### DIFF
--- a/test/gtest/shp/CMakeLists.txt
+++ b/test/gtest/shp/CMakeLists.txt
@@ -8,7 +8,6 @@ add_executable(
   shp-tests
   shp-tests.cpp
   ../common/all.cpp
-  # need to implement same API as MHP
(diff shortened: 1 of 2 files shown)


test/gtest/shp/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ add_executable(
   shp-tests.cpp
   ../common/all.cpp
   # need to implement same API as MHP
```
rscohn2, line 11:
```suggestion
```

And you can probably delete this: https://github.com/oneapi-src/distributed-ranges/blob/de9169185e86f33db4e7dd8ff4a3b86453eecc54/benchmarks/gbench/common/distributed_vector.cpp#L79

PR #318, opened 2023-06-09
Optimize Black-Scholes kernel

BenBrock:
Apply a few common optimizations to Black-Scholes kernels.


BenBrock:
I'm also slightly lowering the size so that oneDPL runs successfully with the default index size (32-bit signed).

A workaround for the odd size performance degradation was added in the last commit, so that is no longer an issue.

PR #324, opened 2023-06-12
Halo exchange if prev != next bug fix

mateuszpn:


include/dr/mhp/halo.hpp
```diff
@@ -363,11 +363,11 @@ class span_halo : public span_halo_impl<T, Memory> {
     drlog.debug(nostd::source_location::current(),
                 "owned groups {}/{} first/last\n", comm.first(), comm.last());
     if (hb.prev > 0 && (hb.periodic || !comm.first())) {
-      owned.emplace_back(span.subspan(hb.prev, hb.prev), comm.prev(),
+      owned.emplace_back(span.subspan(hb.next, hb.next), comm.prev(),
                          communicator::tag::halo_reverse);
     }
     if (hb.next > 0 && (hb.periodic || !comm.last())) {
```
rscohn2, line 356:
```suggestion
    if (hb.prev > 0 && (hb.periodic || !comm.last())) {
```

If next prev is > 0, then we send to next to updates its 'prev' halo

include/dr/mhp/halo.hpp
```diff
@@ -363,11 +363,11 @@ class span_halo : public span_halo_impl<T, Memory> {
     drlog.debug(nostd::source_location::current(),
                 "owned groups {}/{} first/last\n", comm.first(), comm.last());
     if (hb.prev > 0 && (hb.periodic || !comm.first())) {
```
rscohn2, line 352:
```suggestion
    if (hb.next > 0 && (hb.periodic || !comm.first())) {
```

If next halo is > 0, then we send to previous to updates its 'next' halo

include/dr/mhp/containers/distribution.hpp
```diff
@@ -22,6 +22,11 @@ struct distribution {
     return *this;
   }
 
+  distribution &halo(std::size_t prev, std::size_t next) {
```
lslusarczyk, line 25: same signature as previous function, it can not compile (see CI)

also you overwrite here periodic parameter by assigning it false

please remove this function

include/dr/mhp/halo.hpp
lslusarczyk: I understand this suggested change fixes some bug. @mateuszpn , could you please add test cases which fails before and passes after applying this change?

test/gtest/mhp/slide_view-3.cpp
```diff
@@ -13,6 +13,23 @@ TYPED_TEST(Slide3, suite_works_for_3_processes_only) {
...
+TYPED_TEST(Slide3, different_halos) {
+  TypeParam dv(6, dr::mhp::distribution().halo(1, 2));
+  iota(dv, 1);
+  dv.halo().exchange();
+}
+
+TYPED_TEST(Slide3, different_halos_with_0) {
+  TypeParam dv1(6, dr::mhp::distribution().halo(3, 0));
+  iota(dv1, 1);
+  dv1.halo().exchange();
```
lslusarczyk, line 25: Could you please add some checks after exchange (here and in other tests)? Otherwise iota call in this test is not needed. E.g write something to halo region of second process, do exchange and check if new value appeared in the first (or the third) process?

test/gtest/mhp/slide_view-3.cpp
```diff
@@ -13,6 +13,23 @@ TYPED_TEST(Slide3, suite_works_for_3_processes_only) {
...
+
+  TypeParam dv2(6, dr::mhp::distribution().halo(0, 3));
```
lslusarczyk, line 27: Split into separate test cases. The simpler case the better.

lslusarczyk (changes requested):

test/gtest/mhp/halo-3.cpp
```diff
@@ -0,0 +1,65 @@
+#include "xhp-tests.hpp"
+#include <dr/mhp/views/sliding.hpp>
+
+template <typename T> class Halo : public testing::Test {};
+
+TYPED_TEST_SUITE(Halo, AllTypes);
```
lslusarczyk, line 6: Although it is a different binary, but let's make name of suite distinct: Halo3

test/gtest/mhp/halo-3.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TYPED_TEST(Halo, suite_works_for_3_processes_only) {
+  EXPECT_EQ(dr::mhp::default_comm().size(), 3); // dr-style ignore
```
lslusarczyk, line 9: `.size()` is allowed in tests, remove dr-style ignore

test/gtest/mhp/halo-3.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+TYPED_TEST(Halo, dv_different_halos) {
+  TypeParam dv(6, dr::mhp::distribution().halo(1, 2));
+
+  iota(dv, 1);
+  dv.halo().exchange();
+
+  if (dr::mhp::default_comm().rank() == 0) {
+    EXPECT_EQ(*(dv.begin() + 2), 3); // from node 1
```
lslusarczyk, line 19: all these checks use MPI, not halo regions I suppose

in order to see it create a simple test case named `new_halo_values_are_visible_locally_after_exchange`

1. fill vector with one value (e.g. 7)
2. do call the exchange
3. fill vector with different value (e.g. 8)
4. check halo regions as you do it here and expect 7 there (because exchange was not yet here)
5. then call exchange
6. check halo regions as you do it here and expect 8 there (because now halos are distributed to neighbours)

Such a test should fail in step no 4 because when test uses MPI there will be immediately 8 visible.

Now fix the by adding `local()` call on iterator. E.g. `*(dv.begin()+1).local()` and then your test should start work.

In order to have above test case simplest I advise to use halo(1,1)

After creating such a test fix your existing test cases (add `.local()`)

test/gtest/mhp/halo.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+TYPED_TEST(Halo, dv_different_halos) {
+  TypeParam dv(6, dr::mhp::distribution().halo(1, 2));
+  std::vector<int> v(6);
+
+  iota(dv, 1);
+  std::iota(v.begin(), v.end(), 1);
+  dv.halo().exchange();
+
+  EXPECT_TRUE(equal(dv, v));
```
lslusarczyk, line 16: Test does not check halo. I think you could remove line with `dv.halo().exchange()` and it will still pass. This is test of `iota` currently :)

I don't have any idea about sensible checks to write here which are not covered by `Halo3` suite. If you too, then let's remove this test.

test/gtest/mhp/halo-3.cpp
```diff
@@ -0,0 +1,65 @@
+#include "xhp-tests.hpp"
+#include <dr/mhp/views/sliding.hpp>
```
lslusarczyk, line 2: I see which file you copied to create the test :)

test/gtest/mhp/slide_view-3.cpp
```diff
@@ -13,6 +13,7 @@ TYPED_TEST(Slide3, suite_works_for_3_processes_only) {
   EXPECT_EQ(dr::mhp::default_comm().size(), 3); // dr-style ignore
 }
 
+// more Slide3 tests are comming, all assume that there are 3 mpi processes
```
lslusarczyk, line 16: Thank you for taking care of this comment. Slide view does not support different halo sides now, so no tests can be added. Please remove both comments - old and new one.

lslusarczyk (changes requested):

lslusarczyk (approved):

PR #345, opened 2023-06-21
documentation completed with specification of classes and functions

mateuszpn:


doc/spec/source/containers/containers.rst
```diff
@@ -15,4 +15,10 @@ processes and can be accessed by any process in the team.
    :maxdepth: 1
 
    mhp_distributed_vector
+   mhp_distributed_dense_matrix
+
+   shp_vector
+   shp_device_vector
```
rscohn2, line 21:
```suggestion
```

I believe this is internal and should not be documented.

doc/spec/source/containers/containers.rst
```diff
@@ -15,4 +15,10 @@ processes and can be accessed by any process in the team.
    :maxdepth: 1
 
    mhp_distributed_vector
+   mhp_distributed_dense_matrix
+
+   shp_vector
```
rscohn2, line 20:
```suggestion
```

doc/spec/source/containers/mhp_distributed_dense_matrix.rst
```diff
@@ -0,0 +1,23 @@
...
+
+.. include:: ../include/distributed-ranges.rst
+
+.. _distributed_dense_matrix:
+
+=====================================
+``dr::mhp::distributed_dense_matrix``
+=====================================
+
+Synopsis
```
rscohn2, line 13:
```suggestion
Interface
```

For consistency

doc/spec/source/device_memory/device_ptr.rst
rscohn2, line 1: Should remove documentation for this.

rscohn2 (commented):
I will look again when the document builds. We could delete documentation for everything but distributed containers, views, and algorithms.

doc/spec/source/containers/containers.rst
mateuszpn: If so, shouldn't it be in __detail? @BenBrock , can you give any hints on what classes to include in the documentation?

rscohn2 (commented):
Looks good. Please show at the meeting today and we discuss what should be documented, synopsis vs interface, what to do about shp-only algorithms. We also need to document the views, but it could be a separate commit.

doc/spec/source/views/counted_view.rst
```diff
@@ -0,0 +1,24 @@
...
+.. _counted_view:
+
+=============
+ counted_view
+=============
+
+Interface
+=========
+
+Use `std::ranges::views::counted`_.
```
rscohn2, line 16: We no longer support using std::ranges::views. There is always a `dr::mhp::views` or `dr::shp::views`. This is true for all views. I am not sure how to document it. Maybe enough to change
```suggestion
See `std::ranges::views::counted`_.
```

doc/spec/source/views/iota_view.rst
```diff
@@ -0,0 +1,24 @@
...
+.. _iota_view:
+
+===========
+ iota_view
+===========
+
+Interface
+=========
+
+.. doxygenvariable:: dr::mhp::views::iota
```
rscohn2, line 16: We document the interface, not the implementation. So we should document `dr::views::iota` and refer to that. It would be good if the `dr::views` namespace is not shown

doc/spec/source/views/segmented_view.rst
```diff
@@ -0,0 +1,30 @@
...
+ segmented_view
+===============
+
+Interface
+=========
+
+MHP
+---
+
+.. doxygenclass:: dr::mhp::segmented_view
```
rscohn2, line 20: Not intended for public interface. I guess I should move to __detail

doc/spec/source/views/transform_view.rst
```diff
@@ -0,0 +1,25 @@
...
+.. _transform_view:
+
+===============
+ transform_view
+===============
+
+Interface
+=========
+
+.. doxygenclass:: dr::transform_view
```
rscohn2, line 16: We support `dr::mhp::views::transform` and `dr::shp::views::transform` and want to document that. All the supported views are of the form: `dr::xhp::views::ViewName` We do not want to document the `dr::xhp::views::ViewName_view` because we don't consistently support it and it may change over time.

doc/spec/source/containers/mhp_distributed_dense_matrix.rst
```diff
@@ -13,7 +13,12 @@
 Interface
 =========
 
-Not merged yet
+.. code-block:: cpp
+
+   template <typename T, typename Allocator> class distributed_dense_matrix;
+
+The class will be merged and dokumented soon
```
rscohn2, line 20:
```suggestion
The class will be merged and documented soon
```

PR #349, opened 2023-06-22
Inc scan

lslusarczyk:
inclusive scan algorithm implemented in mhp
also added CUDA support to mhp build


rscohn2:
Please put cuda support in a separate PR.

include/dr/detail/communicator.hpp
```diff
@@ -43,6 +43,12 @@ class communicator {
     MPI_Scatter(src, size, MPI_BYTE, dst, size, MPI_BYTE, root, mpi_comm_);
   }
 
+  template <typename T>
+  void scatter(const std::vector<T> &src, T &dst, std::size_t root) const {
+    assert(src.size() >= size());
+    scatter(src.data(), &dst, sizeof(T), root);
+  }
+
```
rscohn2, line 51:
```  
  void scatter(std::contiguous_range auto &&src, T &dst, std::size_t root) const {
    assert(rng::size(src) >= size());
    scatter(rng::data(src), &dst, sizeof(T), root);
  }
```

or maybe std::span instead of std::vector

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,275 @@
...
+    auto dist = rng::distance(in_segment);
+    if (dist == 0) {
+      continue;
+    }
+
+    auto first = ranges::local(rng::begin(in_segment));
+    auto last = ranges::local(rng::end(in_segment));
+    auto s_first = ranges::local(rng::begin(out_segment));
+
+#ifdef SYCL_LANGUAGE_VERSION
```
rscohn2, line 63: Can you move the #ifdef inside the
```
    if (mhp::use_sycl()) {
```
Matching up the curly braces is confusing

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,275 @@
...
+      rng::distance(local_partial_sums, local_partial_sums_iter);
+
+  OVal *local_partial_sums_scanned =
+      local_partial_sums_count
+          ? default_allocator<OVal>().allocate(local_partial_sums_count)
+          : nullptr;
+
+  if (local_partial_sums_count) {
+#ifdef SYCL_LANGUAGE_VERSION
+    if (mhp::use_sycl())
```
rscohn2, line 107: oneapi coding standard suggests always using `{}`, even when `if/else` has 1 statement. Using `{}` and moving `ifdef` into them would make the code easier to read.

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+#ifdef SYCL_LANGUAGE_VERSION
+  std::vector<sycl::event> events;
+#endif
+
+  using OVal = rng::iter_value_t<O>;
+  // how to get allocator used with input and output range? I assume it a
+  // default one probably we should forbid using non-default ones with
+  // distributed_ranges containers (like distributed_vector)
+  OVal *local_partial_sums =
```
rscohn2, line 44: `default_allocator` allocates on the device. You are performing the operation on the host so you need std::allocator. But even better to just use a std::vector so the destructor will take care of deallocation.

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,275 @@
...
+      oneapi::dpl::experimental::for_each_async(
+          dpl_policy(), local_partial_sums_scanned,
+          local_partial_sums_scanned + local_partial_sums_count,
+          [=](auto &&x) {
+            x = binary_op(sum_of_all_guys_before_my_rank_value, x);
+          })
+          .wait();
+    else
+#endif
+      std::for_each(std::execution::par_unseq, local_partial_sums_scanned,
```
rscohn2, line 160: There is an inclusive scan variant that takes an init value. Won't that work here instead of using for_each?

rscohn2 (commented):
Hard to read because of the if/else {}/#ifdef style. I would like to look through again after the next round of changes.

rscohn2:
Please add a simple benchmark like in ``benchmarks/gbench/common/stream.cpp``

include/dr/detail/communicator.hpp
lslusarczyk: I made it consistent with `void gather(const T &src, std::vector<T> &dst, std::size_t root)` where vector is used, but sure, both can be `std::span`. Done.

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: Moving `ifdef` inside `if` will be a bug. But I've improved readability by extracting small functions with `ifdef`s, so main algorithm is free of them now.

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: Braces - done. `ifdef`s left outside but moved to separate functions.

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: This operation is performed on device. But I've changed raw pointer to `unique_ptr` with deleter, which solves deallocation code smell that was here. I've added a description in comment which lists steps and I've marked what is offloaded.

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: No. I want to add constant to each element in this place. Not to compute partial sum. I've explained it in description of alg in comment before inclusive_scan_impl function.

lslusarczyk:
> Please put cuda support in a separate PR. How much works with CUDA? The internal CI system has a cuda machine that could do testing.

Removed CUDA support. Except quick-test which I find very useful to test my changes during development locally on nvidia GPU.
I don't know how much works with CUDA mhp. Not everything. I see shp adds cuda support despite of fact, that not all in shp compiles with it. If we want to add CUDA compilation for things which currently work with it and enable in CI, then it is not so short task and should be done in other task.

Testing with CUDA is worth doing. I saw cases when PVC accepted my kernel and just gave wrong results which test checks had to catch, while CUDA gave nice ILLEGAL_MEMORY_ACCESS when running kernel.

include/dr/detail/communicator.hpp
rscohn2: I thought there was an implicit conversion from vector to span, but apparently not.

include/dr/mhp/algorithms/inclusive_scan.hpp
rscohn2, line 1: There is only 1 segment, and offloading a prefix scan of one element is much slower than doing it on the host.

include/dr/mhp/algorithms/inclusive_scan.hpp
rscohn2: I understand now.

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,305 @@
...
+// move to sycl_support.hpp if need to reuse
+inline void wait_for_events_and_clear(std::vector<Event> &events) {
+#ifdef SYCL_LANGUAGE_VERSION
+  sycl::event::wait(events);
+#endif
+  events.clear();
+}
+
+// inclusive scan algorithm:
+// let's denote segments as: rank(dot)segment_number
```
rscohn2, line 77: This is not the algorithm that I expected: https://courses.cs.washington.edu/courses/cse332/21sp/lectures/19-ParallelPrefix.pdf

I expected up pass to be partial sums and down pass to be scan. But instead there is scan followed by adding a scalar to a vector. Doesn't that require 25% more memory bandwidth?

benchmarks/gbench/common/distributed_vector.cpp
```diff
@@ -256,3 +256,18 @@ static void Mul_Serial(benchmark::State &state) {
 }
 
 DR_BENCHMARK(Mul_Serial);
+
+static void Inclusive_Scan(benchmark::State &state) {
```
rscohn2, line 260: For comparison, please add a benchmark for a std::inclusive_scan with a parallel execution policy

rscohn2:
> Testing with CUDA is worth doing. I saw cases when PVC accepted my kernel and just gave wrong results which test checks had to catch, while CUDA gave nice ILLEGAL_MEMORY_ACCESS when running kernel.

I will try it out. It would be nice to have it working well enough to use for development.

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: Thanks for the pdf. It will be definitely a nice part of my job to learn this new alg. I basically did what is in shp (with a bit more flexible way).

Could you please write what is the most important case we target our system:
1) number of processes << (much lower) number of elements in one process
or 
2) number of processes ~ number of elements in one process
or
3) number of processes >> number of elements in one process
?

I guess the 1st choice.

rscohn2: 1. is the common case.

I see that you followed the SHP algorithm. I was wondering if there is some issue with using sycl device memory. @BenBrock: any comments?

rscohn2: If there is only 1 segment for the entire distributed vector, then it should do an inclusive scan and not the other work. Otherwise, single rank DR will be 2x slower than standard implementation. Not sure what your implementation does in this case.

For gbench, you need micros that show baseline performance  for a single rank that do not use DR. One for SYCL using DPL, one for CPU using std. Verify that single rank DR matches conventional implementation on 1 rank, and reasonable scalability for multiple ranks with DR.

benchmarks/gbench/common/distributed_vector.cpp
lslusarczyk: added, when running on 1 process on my local GPU, there is no difference
```
./mhp-quick-bench --benchmark_filter=Inclusive_Scan*
-----------------------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------------
Inclusive_Scan/real_time               131 ms          131 ms            5 bytes_per_second=11.4077G/s footprint=1.49012G
Inclusive_Scan_Serial/real_time        131 ms          131 ms            5 bytes_per_second=11.3972G/s footprint=1.49012G
```

include/dr/mhp/algorithms/inclusive_scan.hpp
lslusarczyk: good point, my implementation was slower in this case - I've fixed it. Now I use inclusive_scan with init value from the very beginning when computing first non-empty local sum (segment) and skip for_each for this partial sum (segment).
See fixed most recent version.

rscohn2: You still have the issue for the parallel case. A reduce + scan is faster than a scan + vector add. You are memory bound and the vector add uses 2x memory bandwidth compared to reduce.

lslusarczyk: right, I'm fixing it

lslusarczyk: The issue is fixed. Now I use reduce + scan except case of one-segment where just scan is used.
I'd like to test performance, but I've a problem with running mhp-bench with MPI (see slack discussion).

Code looks good to me and is ready to be re-reviewed. Could you please have a look? (unless you want to wait for performance results first)

lslusarczyk:
During working on this an issue in oneDPL was found and fixed. See: https://github.com/oneapi-src/oneDPL/pull/1014 (and also https://github.com/oneapi-src/oneDPL/pull/1015 which we need in order switch to newer oneDPL)

rscohn2 (approved):
It looks good. Feel free to merge when you are ready.

rscohn2:
> During working on this an issue in oneDPL was found and fixed. See: [oneapi-src/oneDPL#1014](https://github.com/oneapi-src/oneDPL/pull/1014) (and also [oneapi-src/oneDPL#1015](https://github.com/oneapi-src/oneDPL/pull/1015) which we need in order switch to newer oneDPL)

If you need this to merge, you can point the CMakeFiles.txt at your fork.

include/dr/mhp/algorithms/inclusive_scan.hpp
```diff
@@ -0,0 +1,362 @@
...
+#pragma once
+
+#ifdef SYCL_LANGUAGE_VERSION
+#include <oneapi/dpl/async>
+#endif
+
+namespace dr::mhp {
+namespace __detail {
+
+template <typename R> std::size_t index_of_first_nonempty_segment(R &&dr) {
```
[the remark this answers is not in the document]
lslusarczyk: > Have you seen empty segments?

No. I've checked and in my tests I don't see them. Just code later assumes segments are non-empty so I wanted to handle them.

lslusarczyk:
> If you need this to merge, you can point the CMakeFiles.txt at your fork.

Thank you. I know. The same way you as did in https://github.com/oneapi-src/distributed-ranges/pull/73
IMO it is enough now to temporary comment out a few checks in one of my tests, which I did. I hope they will merge my fixes soon. If not - I'll point our oneDPL to private branch with fix.

PR #350, opened 2023-06-26
dr::shp::index moved to dr::index

mateuszpn:
dr::shp::index moved to dr::index, to be used with dr::mhp as well

### DIFF
--- a/include/dr/detail/index.hpp
+++ b/include/dr/detail/index.hpp
@@ -9,7 +9,7 @@
 #include <limits>
 #include <tuple>
 
-namespace dr::shp {
+namespace dr {
 
 namespace {
 template <typename T, std::size_t I, typename U = std::any>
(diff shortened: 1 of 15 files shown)


include/dr/detail/index.hpp
```diff
@@ -9,7 +9,7 @@
 #include <limits>
 #include <tuple>
 
-namespace dr::shp {
+namespace dr {
```
lslusarczyk, line 12: I think we have convention all stuff in "dr/detail" should be in namespace `dr::_detail`, not `dr`. I see some classes in dr/detail obey the rule and some don't, but I would use it for new classes.

@rscohn2 , please correct me if I'm wrong

mateuszpn: dr::index will be used as key type for accessing matrix elements - so it cannot be in __details, because users of matrix classes will use it directly.
Still, maybe we should provide dr::shp::index and dr::mhp::index as synonyms to dr::index, similar to what we do in views.

PR #354, opened 2023-06-27
-fscycl moved to top-level CMakeList.txt

mateuszpn:
Global setting of -fsycl flag if SYCL is enabled.


rscohn2 (approved):
I checked the CI time for icpx builld and it is ok.

PR #363, opened 2023-07-06
Debug Heisenbug issue in `shp::gemm`

BenBrock:


BenBrock:
The `async` version still suffers from the Heisenburg, but I have changed `gemm` to use a `buffered`-only version by default.  The person is slightly worse than the `async` version, but this will have to do until we can figure out what's causing the issue.

I suspect it might be a oneAPI runtime/driver bug, but haven't been able to really diagnose what precisely is going wrong.  MKL produces very occasionally incorrect results when we pull in the tiles asynchronously.  If I look at the contents of the tiles before performing the matrix multiply, the contents are correct, and the bug also goes away.  The bug appears both with and without threading.  If I immediately wait on every communication operation, the bug goes away.

PR #373, opened 2023-07-11
mhp gpu selection: round robin among ranks

rscohn2:
@jngkim Recommended adding this method for selecting GPU's for MHP. Previously, I was assuming that we would always be relying on external mechanisms like GPU pinning or resource managers to select GPU's, but @jngkim explained why that may not always work.

This algorithm should do the right thing for all cases of multi/single GPU, GPU pinning/ no pinning, partitionable devices/non partitionable.

@suyashbakshi: I did not test that it works on a multi-gpu system. Please fix it if you find problems.


rscohn2:
clang is failing because `apt install` is getting a 404. This must be a system failure on github runners so I forced the merge.

PR #374, opened 2023-07-11
fix gpu assignment for mhp

rscohn2:
Fix a logic error in the earlier gpu pinning support. 

I assumed that it was being tested by mhp-bench tests, but we were not running mhp-bench with sycl enabled. Enabled mhp-bench  testing with --sycl. It found the problem. Thanks for @jngkim for tracking down the problem.


rscohn2:
Tests are not passing, but they did not pass before this commit either. Will merge.

PR #377, opened 2023-07-13
use cmake-format and fix all files

rscohn2:


rscohn2:
@lslusarczyk : I have no strong preferences about formatting these files. I just want consistency and avoid commits that change format. I see there is a config if you want to make changes to defaults.

lslusarczyk:
> @lslusarczyk : I have no strong preferences about formatting these files. I just want consistency and avoid commits that change format. I see there is a config if you want to make changes to defaults.

A very good solution to our accidental changes in cmake files made by various IDEs. Thank you for this change. I just propose to have concise arguments lists, see https://github.com/oneapi-src/distributed-ranges/pull/395

PR #391, opened 2023-07-19
communicator extension with alltoall() functions

mateuszpn:
Added alltoall() calling MPI_Alltoall(), alltoallv() calling MPI_Alltoallv(), and minor fixes.


include/dr/detail/communicator.hpp
```diff
@@ -85,28 +85,57 @@ class communicator {
...
+    MPI_Alltoall(sendbuf, count * sizeof(T), MPI_BYTE, recvbuf,
+                 count * sizeof(T), MPI_BYTE, mpi_comm_);
+  }
+
+  template <rng::contiguous_range R>
+  void alltoall(R &sendr, R &recvr, int size) {
+    alltoall(sendr.data(), recvr.data(), size);
+  }
+
+  // In alltoallv() methods the sizes and displacements must be scaled by the
```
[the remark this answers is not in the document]
mateuszpn: Required scaling of every element of counts and displacements vectors. Done

lslusarczyk:
I have reviewed these changes also in https://github.com/oneapi-src/distributed-ranges/pull/399 
Please apply my comments here or there (and close this PR in the latter case)

include/dr/detail/communicator.hpp
```diff
@@ -85,28 +93,62 @@ class communicator {
...
+
+  template <rng::contiguous_range R>
+  void alltoall(R &sendr, R &recvr, std::size_t count) {
+    using T = typename R::value_type;
+    MPI_Alltoall(rng::data(sendr), count * sizeof(T), MPI_BYTE,
+                 rng::data(recvr), count * sizeof(T), MPI_BYTE, mpi_comm_);
+  }
+
+  template <typename T>
+  void alltoallv(T *sendbuf, std::vector<int> &sendcnt,
```
lslusarczyk, line 127: I think vectors and bufs in params could be made `rng::contiguous_range`-s like they are in `alltoall`? We will have then more universal signature.

test/gtest/mhp/communicator.cpp
```diff
@@ -0,0 +1,65 @@
...
+template <typename T> class Communicator : public testing::Test {};
+
+using T = int;
+using DV = dr::mhp::distributed_vector<T>;
+
+TEST(Communicator, Alltoallv) {
+  if (comm_size <= 1)
+    return;
+
+  const std::size_t SIZE = 2;
```
lslusarczyk, line 16: Problems with the test
- too much logic in this test (double for loops, while best there should none) 
- it does not test different values being in send and recv displacements vectors (if you accidentaly swap them in the code, the test will not catch it), 
- does not test different sendsizes and recvsizes (if these vectors are mixed in the code it is not caught by test and when values in these vectors are taken from wrong places it is not caught by test)

I propose to 
- create test in 3-ranks only suite.
- have hardcoded all vectors with some different values (including zeros) (like `[1,2,0,4]`)
- at the end test against hardcoded expected value
In this way you will test everything and it will be easy to test reader to verify final value at the end.

lslusarczyk (changes requested):
Some comments.
Also when doing second round of review please do check if all comments from previous round were applied or at least answered why we should not apply them.

Missing comments from previous review:
- https://github.com/oneapi-src/distributed-ranges/pull/399#discussion_r1270259780 is not fully applied (not to _irecv_)
- https://github.com/oneapi-src/distributed-ranges/pull/399#discussion_r1270273490 is not fully applied (const)
- https://github.com/oneapi-src/distributed-ranges/pull/399#discussion_r1270665403 is not answered :)
- https://github.com/oneapi-src/distributed-ranges/pull/399#discussion_r1270279729 is not applied

include/dr/detail/communicator.hpp
mateuszpn: I tried replacing bufs with ranges - but that caused compilation problems in mhp::sort, so I kept the less elegant, still more universal version. And the vectors are storing parameters (sizes, displacements) for MPI API which requires array of the type int*, just what vector<int>.data() returns - so replacing with another type them makes no sense.

include/dr/detail/communicator.hpp
```diff
@@ -85,28 +93,62 @@ class communicator {
...
+                 std::vector<int> &senddsp, T *recvbuf,
+                 std::vector<int> &recvcnt, std::vector<int> &recvdsp) {
+
+    assert(sendcnt.size() == size_);
+    assert(senddsp.size() == size_);
+    assert(recvcnt.size() == size_);
+    assert(recvdsp.size() == size_);
+
+    std::vector<int> _sendcnt(size());
```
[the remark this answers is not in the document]
mateuszpn: ok

test/gtest/mhp/communicator.cpp
lslusarczyk: Skip applying this comment as it is going to be applied separately in https://github.com/oneapi-src/distributed-ranges/issues/412 by @haichangsi

include/dr/detail/communicator.hpp
```diff
@@ -56,26 +56,34 @@ class communicator {
...
   template <typename T>
   void all_gather(const T &src, std::vector<T> &dst) const {
-    assert(dst.size() >= size());
-    all_gather(&src, dst.data(), sizeof(T));
+    assert(rng::size(dst) >= size_);
+    all_gather(&src, rng::data(dst), 1);
+  }
+
+  template <typename T>
+  void all_gather(std::vector<T> &src, std::vector<T> &dst) const {
```
lslusarczyk, line 84: one of vectors can be const

and vectors could be made `rng::contiguous_range` which should work for `vector` and any other class having data and size CPOs

include/dr/detail/communicator.hpp
lslusarczyk: > compilation problems in mhp::sort

I'd like to see that compilation errors. Let's discuss it on person.

lslusarczyk (commented):
Two comments
- missing const
- changing vectors to contiguous_ranges in two places

include/dr/detail/communicator.hpp
```diff
@@ -85,28 +93,65 @@ class communicator {
...
+
+  template <rng::contiguous_range R>
+  void alltoall(const R &sendr, R &recvr, std::size_t count) {
+    using T = typename R::value_type;
+    MPI_Alltoall(rng::data(sendr), count * sizeof(T), MPI_BYTE,
+                 rng::data(recvr), count * sizeof(T), MPI_BYTE, mpi_comm_);
+  }
+
+  template <typename T>
+  void alltoallv(const T *sendbuf, std::vector<std::size_t> &sendcnt,
```
lslusarczyk, line 128: all vectors can be made const

lslusarczyk (approved):

PR #407, opened 2023-07-25
runner refactored, running many benchmarks by one command

lslusarczyk:


lslusarczyk:
Works well in CI - runs multiple analysis by one command. @rscohn2 , could you please have a look? 
@mateuszpn , please check if all my code is well written

PR #414, opened 2023-07-27
benchmarks plotting

lslusarczyk:
- uses seaborn and pandas as @tkarna advised
- creates 6 column pandas.DataFrame out of all _essential_ informations in json files
- adding new plots in a few lines of code by using right selectors over DataFrame
- 2 examples of plots creating

### DIFF
--- a/benchmarks/gbench/CMakeLists.txt
+++ b/benchmarks/gbench/CMakeLists.txt
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: Intel Corporation
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+  #
+  # Google benchmark
+  #
+  set(BENCHMARK_ENABLE_TESTING off)
+  set(BENCHMARK_ENABLE_WERROR off)
+  FetchContent_Declare(
+    googlebench
+    GIT_REPOSITORY https://github.com/google/benchmark.git
+    GIT_TAG v1.8.0)
+  FetchContent_MakeAvailable(googlebench)
+
+  # mhp is not under ENABLE_SYCL to check bechmarks also compilation in gcc
+  add_subdirectory(mhp)
+
+  if(ENABLE_SYCL)
+    add_subdirectory(shp)
+
+    add_custom_target(bench DEPENDS bench-results)
+    add_custom_command(
+      OUTPUT bench-results
+      COMMAND dr-bench analyse -n 1 -n 2 -n 3 -n 4 -s 1000000 -m mhp_gpu -m
+              mhp_cpu -m shp --fork
+      DEPENDS mhp-bench shp-bench)
+  endif()
+endif()
--- a/benchmarks/gbench/CMakeLists.txt
+++ b/benchmarks/gbench/CMakeLists.txt
@@ -23,8 +23,8 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     add_custom_target(bench DEPENDS bench-results)
     add_custom_command(
       OUTPUT bench-results
-      COMMAND dr-bench analyse -n 1 -n 2 -n 3 -n 4 -s 1000000 -m mhp_gpu -m
-              mhp_cpu -m shp --fork
+      COMMAND dr-bench analyze -n 1 -n 2 -n 3 -n 4 -s 1000000 -m mhp_gpu -m
+              mhp_cpu -m shp
       DEPENDS mhp-bench shp-bench)
   endif()
 endif()
--- a/benchmarks/gbench/mhp/mhp-bench.cpp
+++ b/benchmarks/gbench/mhp/mhp-bench.cpp
@@ -63,7 +63,7 @@ int main(int argc, char *argv[]) {
         assert(rank < 1000);
         char tmp[4];
         ::sprintf(tmp, "%03u", rank);
-        ::strncpy(argv[i] + nnnPos, tmp, 3);
+        ::memcpy(argv[i] + nnnPos, tmp, 3);
       }
     }
   }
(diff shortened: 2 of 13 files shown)


benchmarks/gbench/mhp/mhp-bench.cpp
```diff
@@ -54,6 +54,20 @@ int main(int argc, char *argv[]) {
   comm_rank = rank;
   ranks = size;
 
+  // substitute NNN to be rank in output file
+  for (int i = 0; i < argc; ++i) {
+    auto param = std::string(argv[i]);
+    if (param.starts_with("--benchmark_out=")) {
+      const auto nnnPos = param.find("NNN");
```
rscohn2, line 61: This was the recommendation to handle multiple processes: https://github.com/oneapi-src/distributed-ranges/blob/3a7e6973fa2ec185e12001d498daa416b07b9315/benchmarks/gbench/mhp/mhp-bench.cpp#L126
Does it still try to write the output file?

src-python/drbench/drbench/drbench.py
```diff
@@ -45,6 +56,7 @@ def choice_to_mode(c):
 
 
 @cli.command()
+@click.option('--dontplot', is_flag=True, help='just create json files')
```
rscohn2, line 59:
```suggestion
@click.option('--plot', is_flag=True, default=True, show_default=True, help='generate plots')
```

benchmarks/gbench/mhp/mhp-bench.cpp
rscohn2: It is good to have all the ranks, at least for reference. The null reporter stops the console output from the other ranks, which is needed.  The MPI recommendation was to look at the time across all ranks and take the maximum, but that did not seem necessary because we have barriers.

lslusarczyk: Cool. So this is exactly what is implemented here. Jsons are available for all ranks, but plots use just rank 0 numbers.

rscohn2: Yes. If we see strange numbers, we can check if one of the ranks was slower than the rest.

src-python/drbench/drbench/drbench.py
lslusarczyk: My intention was that common use case is to run analysis and plotting together, so extra option `--dontplot` is added occasionally.

I understand you want the user be forced to write '--plot' when he/she wants plots to be generated just after analysis. Adding `default=True, show_default=True` is noop, has the same effect as not writing it.
Please confirm this is what you want and I will change `--dontplot` to `--plot`.

rscohn2: The behavior I want is:
```
dr-bench
```
does the plot and:
```
dr-bench --plot=false
```
Does not plot. It is not common to have negative flags, usually default=true and explicitly disabled. I did not check if `--plot=false` is the proper way to disable a `default=True` flag. The `show_default=True` puts the default value in the help message

rscohn2: I was wrong. Here is the proper way to do it:
https://stackoverflow.com/a/76182645/2525421

lslusarczyk:
@rscohn2 , I think it is ready. I've applied `--no-plot` option support. Also additoinally I reorganized a bit benchmarking cmakelists, so you may want to check it. If all is OK please approve.

benchmarks/gbench/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  # mhp is not under ENABLE_SYCL to check bechmarks also compilation in gcc
+  add_subdirectory(mhp)
+
+  if(ENABLE_SYCL)
+    add_subdirectory(shp)
+
+    add_custom_target(bench DEPENDS bench-results)
+    add_custom_command(
+      OUTPUT bench-results
+      COMMAND dr-bench analyse -n 1 -n 2 -n 3 -n 4 -s 1000000 -m mhp_gpu -m
```
rscohn2, line 26:
```suggestion
      COMMAND dr-bench analyze -n 1 -n 2 -n 3 -n 4 -s 1000000 -m mhp_gpu -m
```

US spelling instead of British

src-python/drbench/drbench/plotter.py
```diff
@@ -2,8 +2,140 @@
...
+    #           mode  vsize          test  nprocs      rtime            bw
+    # 0   MHP_NOSYCL  20000   Stream_Copy       1   0.234987  1.361779e+11
+    # 1   MHP_NOSYCL  20000  Stream_Scale       1   0.240879  1.328468e+11
+    # 2   MHP_NOSYCL  20000    Stream_Add       1   0.329298  1.457645e+11
+    # ..         ...    ...           ...     ...        ...           ...
+    # 62     MHP_GPU  40000    Stream_Add       4  21.716973  4.420506e+09
+    # 63     MHP_GPU  40000  Stream_Triad       4  21.714421  4.421025e+09
+    def __init__(self, plotting_config: PlottingConfig):
+        rows = []
+        for fname in os.listdir('.'):
```
rscohn2, line 68:
```suggestion
        for fname in glob.glob('dr-bench*.rank'):
```
It's a unix-style file path RE. https://docs.python.org/3/library/glob.html
instead of __is_our_file
Not a big deal if you prefer code instead.

rscohn2 (approved):
It looks good.

src-python/drbench/drbench/plotter.py
lslusarczyk: ok for using 'glob', not ok for `.rank` inside glob because shp output does not have `rank`, so it has to be `__is_our_file` in order to not have very complex regexp

benchmarks/gbench/CMakeLists.txt
lslusarczyk: :)

PR #420, opened 2023-07-29
daily ci

rscohn2:


.github/workflows/daily.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: srun -p pvc-shared scripts/devcloud-build.sh
+    - name: Test
+      run: srun -p pvc-shared scripts/devcloud-run.sh
+    - name: Benchmark
+      run: srun -p pvc scripts/devcloud-benchmarks.sh
+    - uses: actions/upload-artifact@v3
+      with:
+        name: log-devcloud-bench
+        path: |
+          build/benchmarks/gbench/mhp/dr-bench*.json
```
lslusarczyk, line 41: `build/benchmarks/gbench/*.png` is missing
`mhp` is in the path no more

The core problem is that the code here and in `pr.yml` to run benchmarks is a multiline copy paste, while it should differ only in one parameter (`pvc` vs `pvc-shared` in scripts/devcloud-benchmarks.sh). Unfortunately yaml does not have any include technique to share code. Maybe it could be shared by some action with parameter where to run benchmarks or you know/have idea about some other way?

README.rst
```diff
@@ -75,21 +44,25 @@ Activate virtual environment::
...
+Build and test with ipcx for SYCL && CPU::
 
-  CXX=clang++ cmake -B build -DENABLE_SYCL=ON
+  CXX=icpx cmake -B build -DENABLE_SYCL=ON
 
-Enable SYCL-MPI examples::
+Build and test with ipcx for SYCL && CPU on devcloud::
 
-  CXX=clang++ cmake -B build -DENABLE_SYCL_MPI=ON
+  CXX=icpx cmake -B build -DENABLE_SYCL=ON -DENABLE_MPIFORK=on
```
lslusarczyk, line 65: If `mpifork` is going to be our default way of running tests (on Intel MPI where as I understand this option is supported) then maybe we should made `MPIFORK=on` as default (if intel MPI is detected) and disabling it should require passing something to cmake. Or not provide any option to cmake, just use fork in tests if intel MPI is found.

I'm afraid about requiring user to pass many options to have our code running in the default way.

rscohn2: mpifork is a workaround for a problem on devcloud. It prevents you from doing multi-node so it is not suitable as a default.

PR #432, opened 2023-07-31
workaround daily scripting bug

rscohn2:
@lslusarczyk

Disabling selective build so I can make progress.


lslusarczyk:
> Disabling selective build so I can make progress.

fixed in https://github.com/oneapi-src/distributed-ranges/pull/434

PR #437, opened 2023-08-01
DotProduct, Black_Scholes and Reduce bw benchmarks

haichangsi:


rscohn2:
Incorporated into other PRs

PR #484, opened 2023-08-15
wrap std::X_iterator so we can use the range-v3 version

rscohn2:


rscohn2:
Abandoning this change. Using rng::forward_range caused compile errors I could not understand. Without that, this change is not useful.

For now, I am working around the original problem by not using iotas that require 64 bits. The larger type causes  range-v3 to use diffmax as a type, which std::forward_range does not understand. A long term fix is to move to gcc ranges library.

PR #488, opened 2023-08-16
Added markers and colors to plots

haichangsi:


rscohn2:
Incorporated into other PR's

PR #526, opened 2023-08-29
Add wave equation statistics

tkarna:
- adds read/write/flop statistics to wave equation
- add those to the benchmark stats too
- fail correctly if L2 error is NaN


rscohn2 (approved):
Looks good. Good timing, I was looking how to report flops for another benchmark.

PR #529, opened 2023-08-30
wave_equation parallel init/checking

rscohn2:
Init and error checking were slow because they did element-wise reads on remote data. Change it to be parallel where ranks work on local data.


tkarna (approved):
Thanks, looks good!

PR #533, opened 2023-08-30
Update mhp to automatically initialize MPI if not already initialized

BenBrock:


include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -128,12 +128,14 @@ class distributed_vector {
   }
 
   ~distributed_vector() {
-    fence();
-    active_wins().erase(win_.mpi_win());
-    win_.free();
-    allocator_.deallocate(data_, data_size_);
-    data_ = nullptr;
-    delete halo_;
+    if (!finalized()) {
```
rscohn2, line 131: A destructor should clean up the resources. For example the unit tests create many distributed_vectors.

rscohn2: Nevermind, I got the logic wrong. I see that this change makes it possible to put the distributed_vector in the same scope as the finalize.

include/dr/mhp/global.hpp
```diff
@@ -145,6 +161,20 @@ inline sycl::queue select_queue(MPI_Comm comm = MPI_COMM_WORLD) {
       devices[dr::communicator(comm).rank() % rng::size(devices)]);
 }
 
+inline void init(sycl::queue q) {
+  __detail::initialize_mpi();
```
BenBrock, line 165: My one reservation about this PR is this line.  According to the MPI spec, we technically don't have a guarantee that all the processes exist until *after* `MPI_Init` is called.  This means that `sycl::queue` might not be valid on all processes after we call `initialize_mpi` here.

However, for most MPI implementations, the processes *will* exist before `MPI_Init` is called, so they will each have their corresponding queues and this code will run fine.

The option would be to remove this line `__detail::initialize_mpi()` for `init(sycl::queue q)`, which would require users to always call `MPI_Init` themselves if they pass in a queue.

rscohn2 (approved):
Looks good

PR #515, opened 2023-08-25
added a gemm benchmark

haichangsi:

PR #536, opened 2023-08-31
Gemm mkl

rscohn2:
Moved #515 here so I can update it.

I see this error if I make `m, n, k` bigger than 1000:
```
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Provided range is out of integer limits. Pass `-fno-sycl-id-queries-fit-in-int' to disable range check. -30 (PI_ERROR_INV\
ALID_VALUE)
```
@BenBrock: Have you seen this? I don't understand why a 2k x 2k matrix causes overflow. I see that shp uses a custom gemm and not MKL. Was there a problem?

### DIFF
--- a/benchmarks/gbench/common/dr_bench.hpp
+++ b/benchmarks/gbench/common/dr_bench.hpp
@@ -100,17 +100,26 @@ namespace xhp = dr::shp;
 class Stats {
 public:
   Stats(benchmark::State &state, std::size_t bytes_read = 0,
-        std::size_t bytes_written = 0)
+        std::size_t bytes_written = 0, std::size_t flops = 0)
       : state_(state) {
     bytes_read_ = bytes_read;
     bytes_written_ = bytes_written;
+    flops_ = flops;
   }
 
   ~Stats() {
-    state_.SetBytesProcessed(reps_ * (bytes_read_ + bytes_written_));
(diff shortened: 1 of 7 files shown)


BenBrock:
`shp` is set up so that you can compile with or without MKL. The CMake actually probably needs to be updated to ensure you're using MKL.  To use MKL, define `-DUSE_MKL` during compilation.  Otherwise, it will use a custom GEMM that is functionally correct, but probably not very fast.  It's probably my GEMM that has the overflow.  I'll check on it.

rscohn2:
I pushed the changes to compile with MKL. There is a use of a deprecated MKL API that is causing the compile to fail.

rscohn2:
```
/localdisk/work/rscohn1/projects/dds/dr/include/./dr/shp/algorithms/matrix/local_gemv.hpp:68:24: error: 'set_csr_data' is deprecated\
:
      Use oneapi::mkl::sparse::set_csr_data(queue, handle, ..., dependencies) instead. [-Werror,-Wdeprecated-declarations]
  oneapi::mkl::sparse::set_csr_data(a_handle, a.shape()[0], a.shape()[1],
```

rscohn2:
The out or range for integer error went away when I added the `-DUSE_MKL` so `custom_gemm` is not suitable for regular use.

PR #544, opened 2023-09-06
fixed sort benchmarks, shp uses its own specific get_queue in benchmarks

lslusarczyk:
measuring only sort, not preparing data for it

### DIFF
--- a/benchmarks/gbench/common/dr_bench.hpp
+++ b/benchmarks/gbench/common/dr_bench.hpp
@@ -36,13 +36,23 @@ extern bool weak_scaling;
 #define DR_BENCHMARK_BASE(x)                                                   \
   BENCHMARK(x)->UseRealTime()->Unit(benchmark::kMillisecond)
 
-#ifdef SYCL_LANGUAGE_VERSION
+#define DR_BENCHMARK_REGISTER_F(fixture, case)                                 \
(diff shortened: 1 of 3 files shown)


rscohn2:
The fixtures looks nice. All the time was spent in the memcpy?

rscohn2 (approved):

lslusarczyk:
> The fixtures looks nice. All the time was spent in the memcpy?

No. I wanted first to move everything except sort outside measurement.
Still on one of our hosts two tests took 20sec regardless of that change and regardless of vector size. It looked as some JIT, but I was unable to remove it neither by eager-jit compile flag nor by calling same function before measurements.

I noticed that this 20sec lag disappears if use default queue from shp instead of queue which is returned by get_queue() from dr_bench.hpp https://github.com/oneapi-src/distributed-ranges/blob/main/benchmarks/gbench/common/dr_bench.hpp#L47

AFAIR  @BenBrock noticed experimentally that having queue/policy created once in process gives better performance that getting them every time from SYCL.

Two conclusions:
- I may create minimal reproducible example and pass it to oneDPL team to investigate.
- we could change mhp/mhp_bench to use the same way of creating queue/policy.



> The fixtures looks nice. All the time was spent in the memcpy?



The fixtures looks nice. All the time was spent in the memcpy?

mateuszpn (approved):

PR #580, opened 2023-10-13
add reduce-max benchmark

suyashbakshi:


rscohn2:
It looks like you started with dev/allocator instead of main. It includes all the dev/allocator commits.

rscohn2:
Can you close this issue and create a new PR based on main branch?

PR #581, opened 2023-10-13
added the codespell hook to pre-commmit

haichangsi:


rscohn2 (approved):
@lslusarczyk: Feel free to merge when you think this is ready.

PR #602, opened 2023-10-27
add for_each with indexing initialization for wave_eq and shallow water

suyashbakshi:


benchmarks/gbench/mhp/shallow_water.cpp
```diff
@@ -875,7 +847,19 @@ int run(
   dr::mhp::fill(ke, 0.0);
 
   // set bathymetry
-  set_field(h, bathymetry, grid, grid.dx / 2, grid.dy / 2, 1);
+  auto h_init_op = [grid, x_offset = grid.dx / 2, y_offset = grid.dy / 2,
+                    row_offset = 1](auto index, auto v) {
```
rscohn2, line 851:
```suggestion
                    row_offset = std::size_t(1)](auto index, auto v) {
```

I don't see a way to specify the type so you have to change the rhs to fix the g++ compile errors.

rscohn2 (approved):

PR #627, opened 2023-11-07
ishmem basic example working on borealis

lslusarczyk:


lslusarczyk:
With no distributed_vector changes yet, but shows working MPI/DR initialization with ISHMEM code.
Using as standalone exec works, using by google test fails.

PR #640, opened 2023-11-15
added new copy signatures for shp and mhp

haichangsi:
remove ifdef BENCH_SHP in sort

### DIFF
--- a/benchmarks/gbench/common/sort.cpp
+++ b/benchmarks/gbench/common/sort.cpp
@@ -22,9 +22,9 @@ public:
     std::vector<T> local(default_vector_size);
     fill_random(local);
     // #ifdef BENCH_SHP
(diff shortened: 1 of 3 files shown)


mateuszpn (changes requested):
That's not exactly what we wanted. In std::ranges, and in distributed ranges, most APIs accept range variable as an argument. There is no need to provide both begin() and end() if we want to make an operation on the whole range (there should be still the option in overload to provide begin() and end()).
Adding the new API with the arguments 'first' and 'last' to mhp::copy() is nice and useful, keep it. However, we need also change in opposite direction: `shp::copy(range in, iterator out)`. The example should read: `xhp::copy(local, rng::begin(*a));` both in mhp and shp.

mateuszpn (commented):
Please add tests for new API

PR #641, opened 2023-11-15
mhp::distributed_vector made customizable with different backends: mpi and ishmem

lslusarczyk:
Distributed Vector made basically customizable. It will be subject for future refactors and updates while enabling new functionality. But the idea could be now seen with this first code illustrating the idea with having template param with backend.

Also added DRLOG macro and swapped message with location placement - I found it more readable to have message at the beginning and then (if I need it) look where the log comes from
sample:
```
initializing MPI skipped - already initialized <global.hpp:77>
calling ishmem_init() <global.hpp:81>
win create:: size: 0 data:0x0 <communicator.hpp:199>
Rank: 1
Running getAndPut test <distributed_vector.cpp:30>
called MPI allocate(20) -> got:0x155481200000 <distributed_vector.hpp:20>
win create:: size: 20 data:0x155481200000 <communicator.hpp:199>
```
Please decide how it looks like to you.


rscohn2:
The refactoring looks ok. The template parameter is an improvement over ifdef's.

For logging, the source file first format was intended to be support IDE's that know how to parse compiler error messages and jump to the code. I didn't find the source lines that useful so moving them to the end is ok.

At one time, I found it useful to use template tests that could test multiple types (int/float for distributed vector) Now I don't think it is worth the extra compilation time and may remove it from the code.

PR #685, opened 2024-01-22
change tpi-ubuntu-latest back to github standard runners

lslusarczyk:


.github/workflows/pr.yml
```diff
@@ -18,7 +18,7 @@ env:
 
 jobs:
   checks:
-    runs-on: cpu
+    runs-on: pvc
```
rscohn2, line 21:
```suggestion
    runs-on: intel-ubuntu-latest
```

.github/workflows/pr.yml
```diff
@@ -133,7 +133,7 @@ jobs:
 
   publish:
     needs: [checks, unit_tests, pvc_unit_tests, ishmem_unit_tests]
-    runs-on: cpu
+    runs-on: pvc
```
rscohn2, line 136:
```suggestion
    runs-on: intel-ubuntu-latest
```

Just run the unit tests on the cluster. These other jobs are ok for the regular public resources

PR #691, opened 2024-01-25
interleave distribution for a distributed vector

haichangsi:


test/gtest/mhp/distributed_vector.cpp
```diff
@@ -113,3 +113,87 @@ TEST(MhpTests, DistributedVectorGranularity) {
...
+    rng::copy(ref, dv.begin());
+    EXPECT_TRUE(equal(dv, ref));
+
+    std::iota(dv.begin(), dv.end(), 31);
+    rng::copy(dv, ref.begin());
+    EXPECT_TRUE(equal(dv, ref));
+  }
+}
+
+TEST(MhpTests, DistributedVectorSegmentSizeCompare) {
```
lslusarczyk, line 182: nice test, give it a better name, e.g. DistributedVector_DifferentSegmentSize_ContentsAreTheSame

You may read sth about names e.g. https://www.softwaretestingmagazine.com/knowledge/how-to-choose-the-right-name-for-unit-tests/ or something else.

lslusarczyk (commented):
really interested test would be in distributed-vector-3 wherte number of processes is fixed and then you create e.g. DV of size 30 with segment size eq 4, call iota, and check local segments on each rank, how many of them you have, how long they are and what are their values

lslusarczyk:
Work very outdated and far from being a complete work. Next attempt will be probably starting this from scratch.

PR #697, opened 2024-01-31
removed unused weak-scaling parameter from drbench suite command

lslusarczyk:


src-python/drbench/drbench/drbench.py
```diff
@@ -247,7 +245,6 @@ def run(
 @option_shp_bench
 @option_dry_run
 @option_clean
-@option_weak_scaling
```
rscohn2, line 250: arguments to suite() below need to be kept in sync with the click @options decorators. I think it is fixed now. Unfortunately no PR testing for dr-bench.

lslusarczyk: Thank you for noticing it. 

Unfortunately it silently passed while I was running manually changed code. Also code was passing for a long time with wrong placement of `different_devices`, so PR testing will not help unless we will add some python UT's which will fail in case of wrong number of parameters or their wrong placement.

PR #707, opened 2024-02-08
DRA-89 fixed (Bug in exclusive_scan for input ranges of small size)

mateuszpn:
For small input vectors  (smaller than  2 * size(r)) sequential function is bug-safe.

### DIFF
--- a/include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
+++ b/include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
@@ -59,8 +59,8 @@ auto inclusive_exclusive_scan_impl_(R &&r, O &&d_first, BinaryOp &&binary_op,
   using value_type = U;
   assert(aligned(r, d_first));
 
-  auto comm = default_comm();
   bool use_sycl = mhp::use_sycl();
+  auto comm = default_comm();
 
   if (rng::size(r) <= 2 * comm.size()) {
     std::vector<value_type> vec_in(rng::size(r));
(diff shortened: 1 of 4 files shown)


lslusarczyk:
Please fix CI, please ensure that UTs do not test this workarounded path of code instead of testing real complex code next (e.g. by adding some log and checking manually that it is not printed out)

include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
```diff
@@ -57,8 +59,32 @@ auto inclusive_exclusive_scan_impl_(R &&r, O &&d_first, BinaryOp &&binary_op,
   using value_type = U;
   assert(aligned(r, d_first));
 
-  bool use_sycl = mhp::use_sycl();
   auto comm = default_comm();
+  bool use_sycl = mhp::use_sycl();
+
+  if (rng::size(r) <= 2 * comm.size()) {
```
lslusarczyk, line 65: add an UT testing this path

include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
```diff
@@ -57,8 +59,32 @@ auto inclusive_exclusive_scan_impl_(R &&r, O &&d_first, BinaryOp &&binary_op,
...
+
+    if (comm.rank() == 0) {
+      if constexpr (is_exclusive) {
+        assert(init.has_value());
+        std::exclusive_scan(detail::direct_iterator(vec_in.begin()),
+                            detail::direct_iterator(vec_in.end()),
+                            detail::direct_iterator(vec_out.begin()),
+                            init.value(), binary_op);
+      } else {
+        std::inclusive_scan(detail::direct_iterator(vec_in.begin()),
```
lslusarczyk, line 78: initt is ignored but it can be non-empty
add 3 test cases for this code: exclusive scan with nonzero init, inclusive scan without init, inclusive scan with init

include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
```diff
@@ -57,8 +59,32 @@ auto inclusive_exclusive_scan_impl_(R &&r, O &&d_first, BinaryOp &&binary_op,
   using value_type = U;
   assert(aligned(r, d_first));
 
-  bool use_sycl = mhp::use_sycl();
   auto comm = default_comm();
+  bool use_sycl = mhp::use_sycl();
```
lslusarczyk, line 63: why 'use_sycl' moved line down?
if there is no reason, then let's don't change code to have easier history

include/dr/mhp/algorithms/inclusive_exclusive_scan_impl.hpp
mateuszpn: Done

mateuszpn (commented):
> Please fix CI, please ensure that UTs do not test this workarounded path of code instead of testing real complex code next (e.g. by adding some log and checking manually that it is not printed out) 

The treshold size, to ensure all segments' size is > 1 is comm.size * (comm.size-1) + 1. Size of test vectors is adjusted to it.

PR #728, opened 2024-02-29
Support for input vectors bigger that 2GB

mateuszpn:
mhp::copy and mhp::sort updated to avoid 32-bit int limitations

### DIFF
--- a/include/dr/detail/communicator.hpp
+++ b/include/dr/detail/communicator.hpp
@@ -207,9 +207,10 @@ public:
     rng::transform(recvdsp, _recvdsp.begin(),
                    [](auto e) { return e * sizeof(valT); });
 
-    MPI_Alltoallv_c(rng::data(sendbuf), rng::data(_sendcnt), rng::data(_senddsp),
-                  MPI_BYTE, rng::data(recvbuf), rng::data(_recvcnt),
-                  rng::data(_recvdsp), MPI_BYTE, mpi_comm_);
+    MPI_Alltoallv_c(rng::data(sendbuf), rng::data(_sendcnt),
+                    rng::data(_senddsp), MPI_BYTE, rng::data(recvbuf),
+                    rng::data(_recvcnt), rng::data(_recvdsp), MPI_BYTE,
+                    mpi_comm_);
   }
 
   bool operator==(const communicator &other) const {
@@ -254,7 +255,8 @@ public:
            std::size_t disp) const {
     DRLOG("MPI comm get:: ({}:{}:{})", rank, disp, size);
     MPI_Request request;
-#if (MPI_VERSION >= 4) || (defined(I_MPI_NUMVERSION) && (I_MPI_NUMVERSION > 20211200000))
+#if (MPI_VERSION >= 4) ||                                                      \
+    (defined(I_MPI_NUMVERSION) && (I_MPI_NUMVERSION > 20211200000))
     MPI_Rget_c(dst, size, MPI_BYTE, rank, disp, size, MPI_BYTE, win_, &request);
 #else
     assert(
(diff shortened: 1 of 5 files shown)


include/dr/detail/communicator.hpp
```diff
@@ -227,7 +228,7 @@ class rma_window {
   void create(communicator comm, void *data, std::size_t size) {
     local_data_ = data;
     communicator_ = comm;
-    DRLOG("win create:: size: {} data:{}", size, data);
+    DRLOG("win create:: size: {} data:{}", (MPI_Aint)size, data);
```
lslusarczyk, line 231: 1. no c-style casts
2. why the cast is needed?

include/dr/detail/communicator.hpp
```diff
@@ -264,12 +273,23 @@ class rma_window {
 
   void put(const void *src, std::size_t size, std::size_t rank,
            std::size_t disp) const {
-    DRLOG("MPI comm put:: ({}:{}:{})", rank, disp, size);
+    DRLOG("MPI comm put:: ({}:{}:{})", rank, disp, (int)size);
```
lslusarczyk, line 276: why cast?

include/dr/detail/communicator.hpp
```diff
@@ -264,12 +273,23 @@ class rma_window {
...
     MPI_Request request;
+
+#if (MPI_VERSION >= 4) ||                                                      \
```
lslusarczyk, line 279: to not repeat long condition please create: `#define MPI_SUPPORTS_RGET_C ...` and use in both places

include/dr/detail/communicator.hpp
```diff
@@ -264,12 +273,23 @@ class rma_window {
...
+#else
+    // MPI_Rput origin_count is 32-bit signed int - check range
+    assert(
+        size <= (std::size_t)INT_MAX &&
+        "MPI API requires origin_count to be positive signed 32-bit integer");
     MPI_Rput(src, size, MPI_BYTE, rank, disp, size, MPI_BYTE, win_, &request);
-    DRLOG("MPI comm wait:: ({}:{}:{})", rank, disp, size);
+#endif
+
+    DRLOG("MPI comm wait:: ({}:{}:{})", rank, disp, (MPI_Aint)size);
```
lslusarczyk, line 290: casts

include/dr/mhp/containers/distributed_vector.hpp
```diff
@@ -38,15 +38,43 @@ class MpiBackend {
...
     DRLOG("calling MPI put(segm_offset:{}, "
           "src:{}, size:{}, peer:{})",
           offset, src, datalen, segment_index);
+
+#if (MPI_VERSION >= 4) ||                                                      \
+    (defined(I_MPI_NUMVERSION) && (I_MPI_NUMVERSION > 20211200000))
+    // 64-bit API inside
     win_.put(src, datalen, segment_index, offset);
+#else
+    for (std::size_t remainder = datalen, off = 0UL; remainder > 0;) {
```
lslusarczyk, line 68: helper function to not repeat code in two places?

include/dr/detail/communicator.hpp
mateuszpn: It was necessary during debug (usigned/signed conversion), now removed,

include/dr/mhp/containers/distributed_vector.hpp
mateuszpn: The difference in the middle (get/put) prevents it. 
In fact, when IMPI 2021.12 will be officially released, and older versions removed from our tests, the code for older MPI should be removed.

PR #729, opened 2024-03-07
Update fft3d.cpp: add timer and FFT3D GFLOPS

jngkim:
Added timer and GFLOPS for the  stand-alone binary for performance analysis.

### DIFF
--- a/benchmarks/gbench/mhp/fft3d.cpp
+++ b/benchmarks/gbench/mhp/fft3d.cpp
@@ -195,8 +195,26 @@ void fft(std::size_t nreps, std::size_t x, std::size_t y, std::size_t z) {
   if (nreps == 0) {
     fft3d.check();
   } else {
+    double elapsed = 0;
     for (int iter = 0; iter < nreps; ++iter) {
+      auto begin = std::chrono::steady_clock::now();
       fft3d.compute();
+      auto end = std::chrono::steady_clock::now();
+      if (iter)
+        elapsed += std::chrono::duration<double>(end - begin).count();
+    }
+
+    if (comm_rank == 0) {
+      std::size_t volume = x * y * z;
+      std::size_t fft_flops =
+          2 * static_cast<std::size_t>(5. * volume *
+                                       std::log2(static_cast<double>(volume)));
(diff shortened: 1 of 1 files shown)


benchmarks/gbench/mhp/fft3d.cpp
```diff
@@ -195,8 +195,26 @@ void fft(std::size_t nreps, std::size_t x, std::size_t y, std::size_t z) {
...
+      auto begin = std::chrono::steady_clock::now();
       fft3d.compute();
+      auto end = std::chrono::steady_clock::now();
+      if (iter)
+        elapsed += std::chrono::duration<double>(end - begin).count();
+    }
+
+    if (comm_rank == 0) {
+      std::size_t volume = x * y * z;
+      std::size_t fft_flops =
```
rscohn2, line 209: Would scaling flops by number of GPUs be the proper way to do weak scaling experiment?

jngkim: I think we should use the total flops for both.

I've modified mhp/fft3d.cpp to run multiple times with --reps and compute FFT3D flops and was planning a PR. I think we can discard this one.

  std::size_t volume = x * y * z;
  std::size_t fft_flops =  2 * static_cast<std::size_t>(5. * volume * std::log2(static_cast<double>(volume)));

  Stats stats(state, 2 * sizeof(real_t) * volume, 4 * sizeof(real_t) * volume, fft_flops);

  distributed_fft<real_t> fft3d(x, y, z);
  fft3d.compute();

  for (auto _ : state) {
    for (std::size_t i = 0; i < default_repetitions; i++) {
      stats.rep();
      fft3d.compute();
    }
  }

jngkim: Sorry, by answer was off.  The size of interest is in BW-limited regime and a fixed volume is fine. In any case, weak-scaling does not make a lot of sense because FFT performance widely varies.

rscohn2: I see, you don't expect that small data size is limiting scalability.

PR #732, opened 2024-03-19
lower_bound support for the find_split_idx func

haichangsi:


haichangsi:
already implemented in #742

PR #740, opened 2024-04-16
shp::sort() - merge instead 2nd sort

mateuszpn:


BenBrock:
Nice catch with line 156, which was almost certainly a memory error.

BenBrock:
Looking at this code again, it seems like we could use a [`parallel_for` instead of a `single_task` here](https://github.com/oneapi-src/distributed-ranges/pull/740/files#diff-d800952cb7b20c6708e84201e50ec49b06afa276bf42314269a7410620639ab4R133), as I originally wrote it.  Do you want to try switching to `parallel_for`?  It should be embarrassingly parallel.

PR #763, opened 2024-05-30
Add ENABLE_OMP_TARGET option to allow OpenMP/SYCL interop.

jngkim:
When ENABLE_OMP_TARGET=ON, OpenMP interop initialize device list to share the context with SYCL.


rscohn2:
This looks good, but it might be better to use cmake to control omp options:
https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html
I expect you would still have to manually specify the target.

jngkim:
Few issues that have to be addressed:
* -fiopenmp -fopenmp-targets=XXX is not handled by cmake. 
* interop/sycl is only available on Intel platforms and ipcx
* Compiler macros _OPENMP does not work due to multiple compilations. At some stages, the macro is reset and results in lost or duplicated kernels.

We should check with our cmake team how we are planning to enable standard CXX + OMP.

PR #786, opened 2024-07-09
Add for_each_n algorithm to shp and mhp

Xewar313:


.gitignore
```diff
@@ -4,3 +4,11 @@
 
 *egg-info
 ._*
+.idea
```
lslusarczyk, line 7: please put developer specific ignores into your private gitignore file
e.g. in your ~/.gitconfig write sth like
```
[core]
        excludesfile = /home/username/.gitignore
```
and in ~/.gitignore write your own rules

include/dr/mhp/algorithms/for_each.hpp
```diff
@@ -53,4 +53,12 @@ void for_each(DI first, DI last, auto op) {
   mhp::for_each(rng::subrange(first, last), op);
 }
 
+/// Collective for_each on iterator/sentinel for a distributed range
+template <dr::distributed_iterator DI, std::integral I>
+void for_each_n(DI first, I n, auto op) {
+  auto last = first;
+  rng::advance(last, n);
```
lslusarczyk, line 60: What happens if `n` is greater than range size? Please add a test for this case.

include/dr/mhp/algorithms/for_each.hpp
```diff
@@ -53,4 +53,12 @@ void for_each(DI first, DI last, auto op) {
   mhp::for_each(rng::subrange(first, last), op);
 }
 
+/// Collective for_each on iterator/sentinel for a distributed range
+template <dr::distributed_iterator DI, std::integral I>
+void for_each_n(DI first, I n, auto op) {
```
lslusarczyk, line 58: std::for_each_n has a return value, please add one too

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -54,4 +54,18 @@ void for_each(Iter begin, Iter end, Fn &&fn) {
   for_each(dr::shp::par_unseq, begin, end, std::forward<Fn>(fn));
 }
 
+template <typename ExecutionPolicy, dr::distributed_iterator Iter,
+          std::integral I, typename Fn>
+void for_each_n(ExecutionPolicy &&policy, Iter begin, I n, Fn &&fn) {
```
lslusarczyk, line 59: std::for_each_n takes Fn by value, mhp too, let's take by value

include/dr/shp/algorithms/for_each.hpp
```diff
@@ -54,4 +54,18 @@ void for_each(Iter begin, Iter end, Fn &&fn) {
...
+  auto end = begin;
+  rng::advance(end, n);
+  for_each(std::forward<ExecutionPolicy>(policy), begin, end,
+           std::forward<Fn>(fn));
+}
+
+template <dr::distributed_iterator R, std::integral I, typename Fn>
```
lslusarczyk, line 66: convention in this file is to name distributed_iterator 'Iter' instead of `R` (used for ranges)

test/gtest/common/for_each.cpp
```diff
@@ -45,6 +45,32 @@ TYPED_TEST(ForEach, RangeAlignedZip) {
   EXPECT_EQ(local, dist);
 }
 
+TYPED_TEST(ForEach, LimitedLength) {
+  Ops1<TypeParam> ops(10);
+
+  auto negate = [](auto &&v) { v = -v; };
+  auto input = ops.vec;
+  auto affected_length = rng::size(input) / 2;
```
lslusarczyk, line 53: cosmetic: the simpler, the better, I would write `5` here

test/gtest/common/for_each.cpp
```diff
@@ -45,6 +45,32 @@ TYPED_TEST(ForEach, RangeAlignedZip) {
...
+
+  auto negate = [](auto &&v) { v = -v; };
+  auto input = ops.vec;
+  auto affected_length = rng::size(input);
+
+  xhp::for_each_n(ops.dist_vec.begin(), affected_length, negate);
+  rng::for_each(ops.vec.begin(), ops.vec.end(), negate);
+
+  EXPECT_TRUE(check_unary_op(input, ops.vec, ops.dist_vec));
+}
```
lslusarczyk, line 72: Please also test with `n` being 0 and `n` exceeding range size.
In order to no duplicate code you may add a helper function `foreach_n_test(uint n)` like sort tests use this [this helper function](https://github.com/oneapi-src/distributed-ranges/blob/main/test/gtest/common/sort.cpp#L12 )

test/gtest/common/for_each.cpp
```diff
@@ -45,6 +45,32 @@ TYPED_TEST(ForEach, RangeAlignedZip) {
   EXPECT_EQ(local, dist);
 }
 
+TYPED_TEST(ForEach, LimitedLength) {
```
lslusarczyk, line 48: Let's have test name ForEachN

lslusarczyk (changes requested):
@Xewar313 congratulations for submitting first PR to DR. looks good. Here are my minor/cosmetic comments.

include/dr/mhp/algorithms/for_each.hpp
lslusarczyk: > What happens if `n` is greater than range size? Please add a test for this case.

agreed it is ub

test/gtest/common/for_each.cpp
lslusarczyk: I meant `TYPED_TEST(ForEachN, LimitedLength)`, ... but let's leave as is

lslusarczyk (approved):

PR #794, opened 2024-07-11
added package task

lslusarczyk:
currently it is empty, added it in order to be able to mark it as required as I still have admin rights this week :)


haichangsi (approved):
lgtm

PR #802, opened 2024-07-16
Add equal to mhp and shp

Xewar313:


lslusarczyk (commented):
Thank you. Especially in SHP it looks perfectly well. Could you please port shp part to oneDPL distributed-ranges branch?

lslusarczyk (approved):

PR #804, opened 2024-07-18
rename:  'mhp' -> 'mp'  and 'shp' -> 'sp'

haichangsi:


lslusarczyk (approved):
Thanks. A very good script.
One comment, we may leave it for future.

PR #812, opened 2024-07-31
added count to mp algorithms

quazuo:


test/gtest/common/count.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+// Fixture
+template <typename T> class Count : public testing::Test {
+protected:
+};
+
+TYPED_TEST_SUITE(Count, AllTypes);
+
+TYPED_TEST(Count, BasicFirstElem) {
+  Ops1<TypeParam> ops(10);
```
mateuszpn, line 15: Fill the vector with values, directly, eg.:
```  
std::vector<int> vec{1, 4, 2, 4, 3, 4, 7, 8, 4, 10, 3, 8, 0, 6, 13, 4};
ops.vec = vec;
xp::copy(ops.vec, ops.dist_vec.begin());
```

include/dr/mp/algorithms/count.hpp
```diff
@@ -0,0 +1,159 @@
...
+template <dr::distributed_range DR> auto count_if(DR &&dr, auto &&pred) {
+  return __detail::count_if(0, false, dr, pred);
+}
+
+//
+// Iterators
+//
+
+// range, elem, w/wo root
+
```
mateuszpn, line 134: Should be implemented as _niebloid_ - compare implementation on [std::ranges::count, std::ranges::count_if](https://en.cppreference.com/w/cpp/algorithm/ranges/count), or implementations within dr, eg. mp::views::enumerate

include/dr/mp/algorithms/count.hpp
```diff
@@ -0,0 +1,159 @@
...
+inline auto add_counts(rng::forward_range auto &&r) {
+  rng::range_difference_t<decltype(r)> zero{};
+
+  return std::accumulate(rng::begin(r), rng::end(r), zero);
+}
+
+inline auto std_count_if(rng::forward_range auto &&r, auto &&pred) {
+  using count_type = rng::range_difference_t<decltype(r)>;
+
+  if (rng::empty(r)) {
```
mateuszpn, line 18: This conditional is redundant in std and dpl version. Please do the check before deciding between std and dpl.
Then you have only single call within std/dpl_count_if - remove the functions, do it inline

include/dr/mp/algorithms/count.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+  if (rng::empty(dr)) {
+    return count_type{};
+  }
+
+  if (aligned(dr)) {
+    dr::drlog.debug("Parallel count\n");
+
+    // Count within the local segments
+    auto count = [=](auto &&r) {
```
mateuszpn, line 57: Consider moving the lambda outside the count_if function

mateuszpn (changes requested):

include/dr/mp/algorithms/count.hpp
quazuo: Some kind of lambda still has to stay here as we need something to pass to `rng::views::transform`. If all of the lambda's code were moved to a separate function, then it would have to take a predicate as a parameter, which would in turn require a new lambda with its only purpose being to pass `pred` to that function. I don't really see a nice solution that would nicely move the lambda outside, unless I'm missing something here

include/dr/mp/algorithms/count.hpp
quazuo: Can you please elaborate on this a bit more? Although I read up a bit on niebloids (as I've never encountered them before), I still don't really get what you mean in this context. What exactly should here be implemented as a niebloid, and more importantly why?

mateuszpn (dismissed):
Thanks for applying my comments. Yet another issues - small in the implementation, more general in tests.

test/gtest/common/count.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+  EXPECT_EQ(std::count_if(ops.vec.begin(), ops.vec.end(), pred),
+            xp::count_if(ops.dist_vec, pred));
+}
+
+TYPED_TEST(Count, FirstElemsIf) {
+  Ops1<TypeParam> ops(20);
+  auto pred = [=](auto &&v) { return v < 5; };
+
+  EXPECT_EQ(std::count_if(ops.vec.begin(), ops.vec.end(), pred),
```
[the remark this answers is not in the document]
quazuo: Thanks, I must have misunderstood the source code for `Ops1` -- fixed it now. I applied changes as per your suggestions and also added another bonus test. I hope all is good now!

include/dr/mp/algorithms/count.hpp
```diff
@@ -0,0 +1,136 @@
...
+template <dr::distributed_range DR>
+auto count_if(std::size_t root, bool root_provided, DR &&dr, auto &&pred) {
+  using count_type = rng::range_difference_t<decltype(dr)>;
+  auto comm = mp::default_comm();
+
+  if (rng::empty(dr)) {
+    return count_type{};
+  }
+
+  if (aligned(dr)) {
```
lslusarczyk, line 42: One range is always aligned with itself. Remove this condition and "else" branch.

lslusarczyk (changes requested):
Mateusz comments applied well.
One more comment. Rest OK.

lslusarczyk (approved):

lslusarczyk:
@quazuo , please merge this change with latest changes on the main, because you need to catch latest changes in CI

lslusarczyk:
Now mp-tests do not link.
```
/usr/bin/ld: /tmp/icpx-d64b1ab94b/all-f158ad.o: in function `oneapi::dpl::__ranges::__select_backend(oneapi::dpl::execution::v1::sequenced_policy)':
all-fdd6ea.cpp:(.text+0x0): multiple definition of `oneapi::dpl::__ranges::__select_backend(oneapi::dpl::execution::v1::sequenced_policy)'; /tmp/icpx-d64b1ab94b/mp-tests-fac318.o:mp-tests-0cb7bc.cpp:(.text+0x0): first defined here
```
Is this problem in our code, your PR, or again in oneDPL?

lslusarczyk:
> Now mp-tests do not link.
> 
> ```
> /usr/bin/ld: /tmp/icpx-d64b1ab94b/all-f158ad.o: in function `oneapi::dpl::__ranges::__select_backend(oneapi::dpl::execution::v1::sequenced_policy)':
> all-fdd6ea.cpp:(.text+0x0): multiple definition of `oneapi::dpl::__ranges::__select_backend(oneapi::dpl::execution::v1::sequenced_policy)'; /tmp/icpx-d64b1ab94b/mp-tests-fac318.o:mp-tests-0cb7bc.cpp:(.text+0x0): first defined here
> ```
> 
> Is this problem in our code, your PR, or again in oneDPL?

Not a problem with this PR but general in DR or oneDPL. The same failure is seen in https://github.com/oneapi-src/distributed-ranges/pull/833

lslusarczyk:
@quazuo , please again merge this change with latest changes on the main. Timeouts were increased and CI should pass on your change now

PR #819, opened 2024-08-05
Add tests and fixes to sp sparse matrix

Xewar313:


include/dr/sp/util/matrix_io.hpp
```diff
@@ -186,23 +186,6 @@ inline coo_matrix<T, I> mmread(std::string file_path, bool one_indexed = true) {
     }
   }
 
-  auto sort_fn = [](const auto &a, const auto &b) {
```
BenBrock, line 189: Sparse matrices *do* need to be sorted after input, since they are not guaranteed to be sorted.  (Typically Matrix Market matrices are sorted by column and then row, the opposite of what we need for CSR.)

Can you explain what's your rationale here?

Xewar313: There is none - I did some experiments aiming to move it to convert_to_csr, but I was unable to do it, and after that I forgot to readd it here. Thanks for catching that!

lslusarczyk (approved):

PR #822, opened 2024-08-16
running CI on CPU and public runners

lslusarczyk:
Temporary change (for ~ 1 week) because of movement of PVC-based CI to new network.

### DIFF
--- a/test/gtest/mp/CMakeLists.txt
+++ b/test/gtest/mp/CMakeLists.txt
@@ -116,7 +116,7 @@ if(ENABLE_SYCL)
     --gtest_filter=-${sycl-exclusions})
   if(NOT MPI_IMPL STREQUAL "openmpi")
     # MPI_Win_create fails for communicator with size 1
-    add_mp_ctest(
+    add_mp_disabled_ctest(
       NAME mp-tests NPROC 1 SYCL TARGS --gtest_filter=-${sycl-exclusions})
   endif()
   # TODO: fix sycl Slide issues, see
--- a/test/gtest/mp/CMakeLists.txt
+++ b/test/gtest/mp/CMakeLists.txt
@@ -107,16 +107,16 @@ if(ENABLE_SYCL)
 
   foreach(nproc RANGE 1 4)
     add_mp_ctest(NAME mp-quick-test NPROC ${nproc} SYCL)
-    add_mp_disabled_ctest(
+    add_mp_pvc_ctest(
       NAME mp-quick-test NPROC ${nproc} OFFLOAD SYCL TARGS --device-memory)
   endforeach()
 
-  add_mp_disabled_ctest(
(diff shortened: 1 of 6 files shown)


test/gtest/mp/CMakeLists.txt
```diff
@@ -107,16 +107,16 @@ if(ENABLE_SYCL)
...
   endforeach()
 
-  add_mp_ctest(
+  add_mp_disabled_ctest(
     NAME mp-tests NPROC 2 TIMEOUT 150 OFFLOAD SYCL TARGS --device-memory
     --gtest_filter=-${sycl-exclusions})
   if(NOT MPI_IMPL STREQUAL "openmpi")
     # MPI_Win_create fails for communicator with size 1
-    add_mp_ctest(
+    add_mp_disabled_ctest(
```
lslusarczyk, line 119: this is not offload test, why disabled?

test/gtest/mp/CMakeLists.txt
```diff
@@ -16,10 +16,10 @@ add_executable(
   ../common/drop.cpp
   ../common/enumerate.cpp
   ../common/equal.cpp
-  ../common/exclusive_scan.cpp
+#  ../common/exclusive_scan.cpp
```
lslusarczyk, line 19: please report issue in JIRA and put its it here and near inclusive scan with some info what is the problem (log snippet with failure, link to failed execution)

test/gtest/mp/CMakeLists.txt
```diff
@@ -107,16 +107,16 @@ if(ENABLE_SYCL)
 
   foreach(nproc RANGE 1 4)
     add_mp_ctest(NAME mp-quick-test NPROC ${nproc} SYCL)
-    add_mp_ctest(
+    add_mp_disabled_ctest(
```
lslusarczyk, line 110: this is a quick workaround

it works for now as we test only on github public runners, however it causes these tests not to run on PVC too if we will want to run them manually

therefore I proposed doing some magic with two separate tags to have set of tests
- good to run on gitghub (all enabled except OFFLOAD ones)
- good to run on PVC (all enabled)

I'd like to have a way to run by hand all previously enabled tests on PVC by one command. Now I can't because good for PVC tests are disabled togerther with e.g. gdb tests which are usually not good to run.

Could you please fix it?
I agree it can be fixed in some next PR after merging this one if we need workaround quickly.

test/gtest/mp/CMakeLists.txt
Xewar313: Done

test/gtest/mp/CMakeLists.txt
Xewar313: Added additional label for running pvc specific tests: PVCLABEL

PR #823, opened 2024-08-16
Split halo into multiple files

capi1500:
Bonus:
new target mp-quick-test for 3 processes only


lslusarczyk:
@capi1500 , please merge your changes with latest main as you need to resolve conflicts and catch latest changes in CI

PR #845, opened 2024-10-09
Fix onedpl inclusion order

Xewar313:


include/dr/mp.hpp
```diff
@@ -6,8 +6,8 @@
 
 #ifdef SYCL_LANGUAGE_VERSION
 #include <oneapi/dpl/algorithm>
-#include <oneapi/dpl/execution>
 #include <oneapi/dpl/numeric>
+#include <oneapi/dpl/execution>
```
lslusarczyk, line 10: please use `// clang-format off` and `// clang-format on` as on this examples:
https://stackoverflow.com/questions/60334299/clang-format-disable-ordering-includes

lslusarczyk (changes requested):

PR #846, opened 2024-10-09
Fix scatter-style local range to distributed range copies

BenBrock:


include/dr/sp/algorithms/copy.hpp
```diff
@@ -120,8 +120,10 @@ sycl::event copy_async(InputIt first, InputIt last, OutputIt d_first) {
     auto local_last = first;
     rng::advance(local_last, n_to_copy);
 
+    auto &&q = __detail::queue(ranges::rank(segment));
+
     events.emplace_back(
-        dr::sp::copy_async(first, local_last, rng::begin(segment)));
+        dr::sp::copy_async(q, first, local_last, rng::begin(segment)));
```
lslusarczyk, line 126: this line does not compile, see CI run
```
/home/runner/work/distributed-ranges/distributed-ranges/include/./dr/sp/algorithms/copy.hpp:126:9: error: no matching function for call to 'copy_async'
  126 |         dr::sp::copy_async(q, first, local_last, rng::begin(segment)));
```

lslusarczyk (changes requested):

PR #860, opened 2024-11-05
Add sparse matrix to mp

Xewar313:


benchmarks/gbench/mp/gemv.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+#else
+
+namespace {
+std::size_t getWidth() {
+  return 8; // default_vector_size / 100000;
+}
+} // namespace
+static auto getMatrix() {
+  std::size_t n = std::max(1., std::sqrt(default_vector_size / 100000)) * 50000;
```
lslusarczyk, line 150: default_vector_size is to be more-less equal to size of data being allocated. However when I run all benchamrks on dnp02 host the only tests which fail with bad_alloc are Gemv_DR.

Please add explanation to the code what are 100000 and 50000 constants for.

Please make the benchamrk allocating more less the same ammount of data like other benchmars so command:\

```
ONEAPI_DEVICE_SELECTOR='level_zero:gpu' I_MPI_OFFLOAD=1 \
I_MPI_OFFLOAD_CELL_LIST=0-11 I_MPI_HYDRA_BOOTSTRAP=ssh \
mpiexec -n 2 -ppn 2  build/benchmarks/gbench/mp/mp-bench --vector-size 1000000000 \
--reps 50 --v=3 --benchmark_filter=GemvEq_DR/ --sycl
```
will not fail on kdse-pre-dnp-02

lslusarczyk (commented):
first few comments, next are comming

benchmarks/gbench/mp/gemv.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+#else
+
+#include "../common/dr_bench.hpp"
+
+#endif
+
+namespace mp = dr::mp;
+
+#ifdef STANDALONE_BENCHMARK
```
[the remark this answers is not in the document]
Xewar313: Fair point, I removed it, since it is not strictly necessary for benchmarking

benchmarks/gbench/mp/gemv.cpp
Xewar313: What's important, it may sometimes freeze - it is due to gtest deciding on some hosts to stop iterating state, and not on others. I am not sure why it decides that, but in that case the test needs to be re-run

benchmarks/gbench/mp/gemv.cpp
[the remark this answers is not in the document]
lslusarczyk: one more STANDALONE_BENCHMARK code is to be removed from this file

test/gtest/mp/reduce.cpp
```diff
@@ -38,16 +38,15 @@ TYPED_TEST(ReduceMP, RootIterators) {
...
-//   auto add = [](auto &&elem) {
-//     return elem + 1;
-//   };
-
-//   auto added = dr::mp::views::transform(r1, add);
-//   auto min = [](double x, double y) { return std::min(x, y); };
-//   auto result = dr::mp::reduce(root, added, 1, min);
-//   EXPECT_EQ(result, 1);
-// }
+TYPED_TEST(ReduceMP, TransformReduce) {
```
lslusarczyk, line 41: great it is fixed
can DRA-192 be closed?

lslusarczyk (commented):
A few comments. I will be happy to see a very minimal example of using one, most useful format of CSR, in our tutorial repo. And we can inform UMCS that they can start using it.

lslusarczyk (approved):

